PR #49, opened 2026-01-16
Patch ibflex pydantic models to ignore extra fields

vroonhof:
### Motivation
- Interactive Brokers/ibflex sometimes adds new fields to Flex XMLs which cause strict pydantic models in `ibflex` to raise parsing errors; the goal is to make parsing tolerant to extra fields. 

### Description
- Added `inspect` and `pydantic.BaseModel` imports and implemented `_patch_ibflex_pydantic_models(module: Any) -> int` which walks the `ibflex` module and updates model configuration to ignore extra fields. 
- The patch sets `model_config["extra"] = "ignore"` for Pydantic v2 models and `Config.extra = "ignore"` for v1-style models, calls `model_rebuild(force=True)` when available, and counts patched models. 
- The patch is executed at import time when `ibflex` is available and logs the number of patched models; changes are in `src/opensteuerauszug/importers/ibkr/ibkr_importer.py`.

### Testing
- Ran `pytest tests/importers/ibkr/test_ibkr_importer.py -k valid` and the selected test passed (`1 passed, 16 deselected`).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_696a706c9c14832e85af49b759c41e24)

### DIFF
--- a/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
+++ b/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
@@ -1,11 +1,14 @@
 import os
 import logging
+import inspect
 from typing import Final, List, Any, Dict, Literal
 from datetime import date, timedelta
 from decimal import Decimal, InvalidOperation
 from collections import defaultdict
 
 logger = logging.getLogger(__name__)
+
+from pydantic import BaseModel
 
 from opensteuerauszug.model.position import SecurityPosition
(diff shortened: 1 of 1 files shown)


vroonhof:
the idea was not run that one test, but instead do / add a test that confirms that this technique works for ignoring unknown fields and causes no other issues.

vroonhof:
now natively supprted in our vendered copy of ibflex.

PR #105, opened 2026-01-29
Skip DA1 fields when tax amounts are zero

vroonhof:
### Motivation
- Prevent emitting DA1 attributes (like `lumpSumTaxCredit` and `additionalWithHoldingTaxUSA`) when computed DA1 tax amounts are zero and ensure DA1 fields are produced as a single bundled unit when present.

### Description
- In `src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py` compute `lump_sum_amount` and `non_recoverable_amount` and only set DA1 percent/amount fields and `lumpSumTaxCredit` when either amount is greater than zero, and set `additionalWithHoldingTaxUSA` for US only inside the same branch.
- Merged previous separate conditionals into the main `if da1_rate` branch so DA1 fields are emitted together rather than separately.
- In `src/opensteuerauszug/calculate/minimal_tax_value.py` only write the default `additionalWithHoldingTaxUSA = 0` when at least one DA1 field is present by adding a `has_da1_fields` check.
- Existing percent/amount calculations and quantization logic for DA1 values were preserved.

### Testing
- No automated tests were run for this change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6979ba1268f0832ebbcca557f4da28be)

### DIFF
--- a/tests/calculate/test_us_withholding_tax_defaults.py
+++ b/tests/calculate/test_us_withholding_tax_defaults.py
@@ -17,7 +17,8 @@ from opensteuerauszug.model.ech0196 import (
 def test_fix_issue_78_us_security_additional_withholding_tax_usa():
     """
     Test that FillInTaxValueCalculator correctly sets additionalWithHoldingTaxUSA to 0
-    for US securities if it is missing, when in FILL mode.
+    for US securities when other DA1 fields are present, if it is missing, when in
+    FILL mode.
     """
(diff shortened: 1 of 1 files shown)


vroonhof:
This now needs manual integration resting if it fixes the issue. Code looks xleaner

vroonhof:
Fixed the warning for my data.

PR #114, opened 2026-01-30
Align PDF417 barcode with spec

pet-zh:
"Beilage zu eCH-0196 V2.2.0 – Barcode Generierung – Technische Wegleitung":
"Zur eindeutigen Identifikation beim Scannen sind in den Metadaten im Attribut PDFMacroFileName die ID (siehe Kapitel 2.1) und im Attribut PDFMacroFileId eine zufällige Zahl (4 Integer) abgelegt."

### DIFF
--- a/src/opensteuerauszug/render/render.py
+++ b/src/opensteuerauszug/render/render.py
@@ -1,7 +1,7 @@
 import io
 from math import floor
 import sys
-import random
+import hashlib
 import os
 import json
 from pathlib import Path
@@ -771,7 +771,7 @@ def render_to_barcodes(tax_statement: TaxStatement) -> list[PILImage.Image]:
     xml = tax_statement.to_xml_bytes()
     data = zlib.compress(xml)
 
-    FILE_NAME = tax_statement.id
+    file_name = tax_statement.id
 
     # Follow Guidance in "Beilage zu eCH-0196 V2.2.0 – Barcode Generierung – Technische Wegleitung"
     # our library foes not allow setting the row_count, so guess by making the segments roughly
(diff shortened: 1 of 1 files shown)


src/opensteuerauszug/render/render.py
```diff
@@ -770,29 +771,35 @@ def render_to_barcodes(tax_statement: TaxStatement) -> list[PILImage.Image]:
...
+    #    1 + 2 + 4 macro pdf fields with 4 word file ID
     #    4 for segment count
     #    1 for possible last code marker
     #    32 error correction at level 4
     #    1 for specifying byte encoding
-    # gives 43 words of overhead
-    FIXED_OVERHEAD = 43
+    # gives 46 words of overhead
+    # plus file name length (it is actually less because of compression, but be safe)
+    FIXED_OVERHEAD = 46 + len(FILE_NAME)
```
vroonhof: now that it is no longer fixed, change name (or add the filename seperately in the final computation.

src/opensteuerauszug/render/render.py
```diff
@@ -770,29 +771,35 @@ def render_to_barcodes(tax_statement: TaxStatement) -> list[PILImage.Image]:
     xml = tax_statement.to_xml_bytes()
     data = zlib.compress(xml)
 
+    FILE_NAME = tax_statement.id
```
vroonhof: don't use ALL_CAPS for things that are not constant.

src/opensteuerauszug/render/render.py
```diff
@@ -770,29 +771,35 @@ def render_to_barcodes(tax_statement: TaxStatement) -> list[PILImage.Image]:
...
+    # plus file name length (it is actually less because of compression, but be safe)
+    FIXED_OVERHEAD = 46 + len(FILE_NAME)
     # Given in the guidance
     NUM_COLUMNS = 13
     NUM_ROWS = 35
     CAPACTITY = NUM_COLUMNS * NUM_ROWS - FIXED_OVERHEAD
     # Byte encodinge efficency is 6 bytes per 5 codewords
     SEGMENT_SIZE = floor((CAPACTITY / 5) * 6)
+    # Official PDF generator uses 4 * 3 digit (<= 255 each)
+    FILE_ID = [random.randint(100, 255) for _ in range(4)]
```
vroonhof: normal variable name.

vroonhof: I actually don't like it being really random. Ideally we would compute this based on some hash of the content. Leave a TODO comment,

vroonhof (changes requested):
Thanks you for sending this.

see inline.

vroonhof:
side comment: I think this suggestion in the spec is fairly wasteful becauset he filename gets repeated in each barcode uncompressed (and the importer is very broken if relies on client supplied input).

pet-zh:
> side comment: I think this suggestion in the spec is fairly wasteful becauset he filename gets repeated in each barcode uncompressed (and the importer is very broken if relies on client supplied input).

Agree on this, but let's align to official standards as much as possible. That reduces the risk of non-importable or rejected reports in the end.

I have addressed your review - thanks.

vroonhof:
FYI this PR broke actual decoding of the barcocdes. See issue #239 (dididn't have time yet to figure out why. I suspect filename handling in the underlying macro-pdf encode). I reverted everything for now while we investigate

PR #120, opened 2026-02-03
Only skip hyphen IBKR account entries

vroonhof:
### Motivation
- Flex statements and entry rows in IBKR exports should only be filtered out when they are explicit pseudo-summary rows marked with `accountId="-"`, and rows that merely omit `accountId` must be kept and processed.

### Description
- Change importer to only skip FlexStatements when `accountId == "-"` and not when `accountId` is missing, by checking `account_id == "-"` when parsing statements.
- Add a scoped `should_skip_entry` check that only treats entries as pseudo when their `accountId == "-"` and apply it to Trades, OpenPositions, Transfers, CorporateActions and CashTransactions.
- Also skip CashReportCurrency entries only when their `accountId == "-"` to avoid excluding rows with missing `accountId`.
- Update tests in `tests/importers/ibkr/test_ibkr_importer.py` to use a fixture with missing `accountId` on some rows and assert that those rows are retained (two securities and two payments), and adjust related transfer fixtures.

### Testing
- Ran the full test suite with `pytest`, resulting in `465 passed, 5 skipped`.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_698208885314832e9ea7f03344c34fec)

### DIFF
--- a/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
+++ b/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
@@ -1,5 +1,6 @@
 import os
 import logging
+import xml.etree.ElementTree as ElementTree
 from typing import Final, List, Any, Dict, Literal, get_args, cast
 from datetime import date, timedelta
 from decimal import Decimal, InvalidOperation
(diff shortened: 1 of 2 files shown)


vroonhof:
superceded by https://github.com/vroonhof/opensteuerauszug/pull/125

PR #142, opened 2026-02-07
Another round style and label adjustments

pet-zh:
This is another step closer to the official layout.

It adjusts colors, spacings, texts and layout.

In addition, it moves OpenSteuerAuszug link to the footer.

### DIFF
--- a/src/opensteuerauszug/render/render.py
+++ b/src/opensteuerauszug/render/render.py
@@ -265,7 +265,7 @@ def create_client_info_table(tax_statement: TaxStatement, styles, box_width: flo
                            Paragraph(f"<b>{escape_html_for_paragraph(client_info['period'])}</b>", info_style)])
     
     if 'created' in client_info:
-        table_data.append([Paragraph("Erstellt am", info_style),
+        table_data.append([Paragraph("Daten von", info_style),
                            Paragraph(escape_html_for_paragraph(client_info['created']), info_style)])
 
     if 'canton' in client_info:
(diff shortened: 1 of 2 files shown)


src/opensteuerauszug/render/render.py
```diff
@@ -312,20 +320,14 @@ def draw_page_header(canvas, doc, is_barcode_page: bool = False):
         if institution_name:
```
vroonhof, line 320: I prefer to keep some indication of Opensteuerauszug close to the institution name. I think the gray line at the bottom is not sufficient.  I though quite long whether to even put the institution there as it isn't published by them. I think we can keep the URL at the bottom to make it look less uggly

src/opensteuerauszug/render/render.py
```diff
@@ -253,25 +253,33 @@ def create_client_info_table(tax_statement: TaxStatement, styles, box_width: flo
...
+
     if 'period' in client_info:
-        table_data.append([Paragraph(f"Periode: <b>{escape_html_for_paragraph(client_info['period'])}</b>", info_style)])
+        table_data.append([Paragraph("Periode", info_style),
+                           Paragraph(f"<b>{escape_html_for_paragraph(client_info['period'])}</b>", info_style)])
     
     if 'created' in client_info:
-        table_data.append([Paragraph(f"Daten von: {escape_html_for_paragraph(client_info['created'])}", info_style)])
-    
+        table_data.append([Paragraph("Erstellt am", info_style),
```
vroonhof: this was another intentional deviation from the official layout. I want our date to be the one the original export from the broker happened.

vroonhof (changes requested):

vroonhof (approved):

PR #144, opened 2026-02-07
Do not render trailing 0 in XML

pet-zh:
There is no value added, and those consume minimal space in barcodes. This also something I have seen in some of my real world statements.

### DIFF
--- a/src/opensteuerauszug/model/ech0196.py
+++ b/src/opensteuerauszug/model/ech0196.py
@@ -420,9 +420,7 @@ class BaseXmlModel(BaseModel):
                         raise NotImplementedError("Base64 encoding is not implemented")
                     elif isinstance(value, Decimal):
                         # Format Decimal as a plain string to avoid scientific notation and remove trailing zeros
-                        normalized = value.normalize()
-                        # Remove trailing .0 if present (e.g., 1.0 -> 1)
-                        str_value = format(normalized, 'f').rstrip('0').rstrip('.') if '.' in format(normalized, 'f') else format(normalized, 'f')
+                        str_value = format(value.normalize(), "f")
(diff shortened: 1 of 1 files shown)


src/opensteuerauszug/model/ech0196.py
```diff
@@ -419,8 +419,10 @@ def _build_attributes(self, element: ET._Element):
                         # TODO: Implement base64 encoding if needed for fileData
                         raise NotImplementedError("Base64 encoding is not implemented")
                     elif isinstance(value, Decimal):
-                        # Format Decimal as a plain string to avoid scientific notation
-                        str_value = f'{value:f}'
+                        # Format Decimal as a plain string to avoid scientific notation and remove trailing zeros
+                        normalized = value.normalize()
+                        # Remove trailing .0 if present (e.g., 1.0 -> 1)
+                        str_value = format(normalized, 'f').rstrip('0').rstrip('.') if '.' in format(normalized, 'f') else format(normalized, 'f')
```
vroonhof: my AI friends think that str_value = format(value.normalize(),"f") is sufficient without the ugly strip and one one example I tried this is indeed true.

pet-zh: @vroonhof Looks like your AI friend is right here - mine (I think it was copilot in this case) somehow wanted to do it in a more extensive way ;) Applied - thanks!

vroonhof:
This now actually fails one of my real data tests 

FAILED tests/model/test_ech0196.py::test_xml_round_trip_files[/workspaces/opensteuerauszug/private/samples/UBS_fixed.xml]

because apparently UBS does not do this. Need to ponder what to do here..

PR #146, opened 2026-02-07
PDF: Ensure stable sorting for accounts, payments and securites

### LINKED ISSUE #93: PDF: Order events (e.g. payments) by event date
The DL example XML does not have ordered all events (e.g. interest payments) by event date. We should order all events by date (e.g. interest, security payments) before printing them into the table.

pet-zh:
This ensures stable sorting in PDF even when XML input is not sorted (which is not required as per spec).

This also allows easier comparison with PDFs render with offical tooling.

Resolves #93

### DIFF
--- a/src/opensteuerauszug/render/render.py
+++ b/src/opensteuerauszug/render/render.py
@@ -1343,6 +1343,12 @@ def create_securities_table(tax_statement, styles, usable_width, security_type: 
         depot_header_rows.append(current_row)
         current_row += 1
 
+        securities_in_depot.sort(
+            key=lambda s: (
+                int(s.valorNumber) if s.valorNumber is not None else 0,
+                s.securityName if s.securityName is not None else ''
+            )
+        )
         for security in securities_in_depot:
             # Description/header row for the security
             if security.country != "CH" and security.country != None:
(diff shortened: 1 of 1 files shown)


src/opensteuerauszug/render/render.py
```diff
@@ -1278,7 +1279,13 @@ def create_securities_table(tax_statement, styles, usable_width, security_type:
     intermediate_total_rows = []
     current_row = 1  # Start after header
 
-    
+    filtered_securities.sort(
```
vroonhof: Surely we must keep the depot as the primary key so that depots stay together.

src/opensteuerauszug/render/render.py
```diff
@@ -1278,7 +1279,13 @@ def create_securities_table(tax_statement, styles, usable_width, security_type:
...
+        key=lambda pair: (
+            pair[1].valorNumber is not None,
```
vroonhof: Is this sort order (items without valor first) the one in the example tool you are comparing with?  If so I think it could be achieved more simply by setting it to zero if missing and using it as the main key.

src/opensteuerauszug/render/render.py
pet-zh: Good catch - done.
Displaying the depot number in the PDF is another TODO, but not addressed here.

pet-zh:
@vroonhof I have resolved the merge conflicts here.

In addition, I have also taken ex-date into account for payments.

PR #147, opened 2026-02-07
Summary table layout spacing fixes

pet-zh:
Added TODOs for splitted up DA-1 and USA values since I am not sure how to get those (only difference compared to the official template).


vroonhof:
Difficult to reviews without running it through a visible example. But i don't want you to wait. Let's merge.

PR #159, opened 2026-02-11
Improve raw-import

### LINKED ISSUE #97: render.py: Running render.py directly causes 0 values for some totals
For testing adjustments in `render.py`, I run this class directly.

This causes for example `0` value for `Steuerwert der A- und B-Werte am 31.12.202X` on summary page.

This value is defined in the XML top level as attribute (`totalTaxValue`).

Some applies to the summary line at the right `(1) Davon A XXX und B XXX`.

Making this work would be very handy to also test any other real world XML files. Or do I miss something here?

pet-zh:
- fill in missing fields to make rendering possible in render phase
- fix DA1 detection in TotalCalculator
- minor other cleanups

Resolves #97

### DIFF
--- a/src/opensteuerauszug/calculate/total.py
+++ b/src/opensteuerauszug/calculate/total.py
@@ -1,6 +1,7 @@
 from decimal import Decimal
 from typing import List, Optional, Dict, Any, cast
 
+from ..core.security import determine_security_type
 from ..model.ech0196 import (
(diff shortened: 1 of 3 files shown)


pet-zh:
With this change, I am almost able to reproduce the [DL sample](https://datalevel.ch/assets/downloads/dl-sampleibkr_echsteuerauszug_stj2024.pdf) (some minor differences need more investigation).

PR #161, opened 2026-02-12
Importer: Do not add 0 quantity start-of-period stock entry

### LINKED ISSUE #136: XML: Do not generate blank initial holdings (`stock`) when balance is zero
Follow-up from https://github.com/vroonhof/opensteuerauszug/issues/130:

The initial holdings line does not need to be printed in case it was 0 (in the below example for 10.4.2025) (`stock` element with `mutation=0` and `quantity=0`):
Image

In my real world examples, there is not even a line in such cases in the XML. So I would suggest to move this check from rendering to XML generation. 

Sorry, my initial comment wasn't fully clear about this.

The spec is clear about this ([p32](https://www.ech.ch/sites/default/files/dosvers/hauptdokument/STAN_d_DEF_2022-06-07_eCH-0196_V2.2.0_E-Steuerauszug_.pdf)):
> Der Bestand (Mutationen) zum Titel. Ein Anfangsbestand zum Periodenbeginn (`periodFrom`) ist anzugeben, wenn dieser grösser 0 ist.
Alle Zugänge und Abgänge innerhalb des Steuerjahres sind als Mutationen anzugeben. Ein
Endbestand zum Periodenende (`periodTo`) ist nicht anzugeben, da sich dieser aus dem Anfangsbestand, den Zugängen und Abgängen im Steuerjahr ergibt.

That means we should apply this change to the imported, and revert the renderer change done in https://github.com/vroonhof/opensteuerauszug/pull/131.

pet-zh:
According this spec, such entries must not be added.

Since we are doing this now in importer lever, handling in renderer can be removed (reverts 3ac9d7f).

Resolves #136

### DIFF
--- a/src/opensteuerauszug/importers/schwab/schwab_importer.py
+++ b/src/opensteuerauszug/importers/schwab/schwab_importer.py
@@ -142,11 +142,6 @@ class SchwabImporter:
             if start_pos_synth:
                 qty_to_set_at_start = start_pos_synth.quantity
                 logger.info(f"[{current_identifier}] Synthesized start position for {self.period_from}: Qty {qty_to_set_at_start} {currency_at_start}")
-            else:
-                earliest_stock_date = min(s.referenceDate for s in live_stocks_list) if live_stocks_list else None
-                if not earliest_stock_date or earliest_stock_date > self.period_from or \
-                   (earliest_stock_date == self.period_from and live_stocks_list[0].mutation):
-                    logger.info(f"[{current_identifier}] No suitable existing/synthesizable start balance for {self.period_from}. Inserting zero balance.")
 
             if qty_to_set_at_start != 0:
                 start_balance_stock = SecurityStock(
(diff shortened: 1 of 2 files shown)


src/opensteuerauszug/importers/schwab/schwab_importer.py
```diff
@@ -148,15 +148,17 @@ def _reconcile_and_ensure_boundary_stocks_for_position(
                    (earliest_stock_date == self.period_from and live_stocks_list[0].mutation):
                     logger.info(f"[{current_identifier}] No suitable existing/synthesizable start balance for {self.period_from}. Inserting zero balance.")
```
vroonhof: this will now never happen, I think that makes the whole else block dead code.

tests/importers/ibkr/test_ibkr_importer.py
```diff
@@ -671,15 +668,7 @@ def test_transfer_with_open_position_no_trades(sample_ibkr_settings):
         assert iwda_sec.country == "IE"
         
         # Check stock entries: should have opening balance, transfer mutation, and closing balance
```
vroonhof: this should now have "should NOT have opening balance (because zero)

tests/importers/ibkr/test_ibkr_importer.py
vroonhof, line 376: maybe add a comment there that this path is now checking the case where opening balance is non-trivial (the tests in this file are horrible, but that is another matter).

vroonhof (approved):
Thanks for the adjustments

vroonhof:
This PR actually broke integration testing on my own data because some downstream logic assumes there is always a position entry.

'''
INFO:opensteuerauszug.calculate.cleanup:Generated new TaxStatement.id: CH19334IBKRU2333563XX2024123101
Error during phase calculate: Could not determine stock quantity for security 'US02072L5654' (Depot: U2333563, Payment: 'BOXX(US02072L5654) CASH DIVIDEND USD 0.16782 PER SHARE (Short Term Capital Gain)' on 2024-08-14, exDate: N/A) using date 2024-08-14 (as paymentDate). Check stock history. Current quantity remains -1.
Stack trace:
'''

PR #182, opened 2026-02-16
IBKR: Improve stock mutations

pet-zh:
- use consistent label for both aggregated and non-aggregated trades
- set `unitPrice` attribute instead of appending trade price into label

### DIFF
--- a/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
+++ b/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
@@ -179,11 +179,11 @@ class IbkrImporter:
                     # test for same sign of quantity
                     and (pending.quantity * stock.quantity) > 0
                 ):
-                    pending.quantity += stock.quantity
-                    if pending.quantity > 0:
-                        pending.name = "Buy"
-                    else:
-                        pending.name = "Sell"
+                    total_quantity = pending.quantity + stock.quantity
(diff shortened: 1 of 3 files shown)


src/opensteuerauszug/importers/ibkr/ibkr_importer.py
```diff
@@ -179,18 +179,19 @@ def _aggregate_stocks_by_date(self, stocks: List[SecurityStock]) -> List[Securit
                     # test for same sign of quantity
                     and (pending.quantity * stock.quantity) > 0
                 ):
-                    pending.quantity += stock.quantity
-                    if pending.quantity > 0:
-                        pending.name = "Buy"
```
vroonhof, line 184: does that mean that in the output we now no longer have "Buy" or "Sell" I feel keeping that is cleaner.

vroonhof (commented):
sorry don't have time check output in my example yet.

src/opensteuerauszug/importers/ibkr/ibkr_importer.py
pet-zh: No, the labels BUY or SELL are still preset. They are set before aggregating. In my understanding this block here got added to override it so that they no longer display for example price in the label, because that would be wrong after aggregation. Since I have cleaned up the initial label, this is no longer necessary in my understanding.

Issue #156, opened 2026-02-08
Consider rendering DA-1 subtotals by country


vroonhof:
My swiss banks do this. I have not checked if the standard/example renderer in taxstatement.jar does as well.

It seems purely cosmetic at this point as it is not in the XML data. (though for it to be useful to compare with for example 1042-S it would be nicer if it also kept totals in local currency, but nobody does and the kursliste doesn't even provided it per se)

pet-zh:
Yes, the official rederer does this.

BTW, I cane across a pretty extensive IBKR example from DL which uses the official renderer: https://datalevel.ch/assets/downloads/dl-sampleibkr_echsteuerauszug_stj2024.pdf

You probably want to add this to your local test suite.

pet-zh:
@vroonhof Please reopen. My PR only adds country headers, but not subtotals.
Subtotals row should contain subtotals for: Steuerwert, Bruttoertrag, Anrechenbare ausl. Quellensteuer, Steuerrückbehalt USA. Left label should be "Total US" where US is the country.

PR #196, opened 2026-02-18
PDF: Print country headers for DA1 securities

pet-zh:
Ordering by country for DA1 securities is also applied by official renderer.

Related to #156


vroonhof:
Can you sync and see if you can trigger the new example pdf renderer..

pet-zh:
@vroonhof Done. Works like a charm - nice work on the PR PDF preview tooling!

Please note that main is currently broken due to a merge race. I have included the fix here.

PR #198, opened 2026-02-18
Config: Enforce users to copy template file

### LINKED ISSUE #113: IBKR: Take canton by default from export without modifying `config.toml`
_Originally posted by @pet-zh in [#77](https://github.com/vroonhof/opensteuerauszug/issues/77#issuecomment-3797235015)_

I would prefer to not have a default value set in `config.toml`. The current solution requires me to edit the file for an IBKR statement.
I would suggest to either remove the default value or handle it like `full_name` (always take it from broker export if available).

pet-zh:
This reduces the chance that personal data is getting committed. In most cases a config file is not needed.

Resolves #113


pet-zh:
@vroonhof Please challenge this approach. There is also other ways to improve this behavior of course.

Issue #197, opened 2026-02-18
B-securities subtotal wrong in case cash bank account holdings are present


pet-zh:
In case B-securities are present, the subtotals at the end of all B-securities (`Total B-Werte`) is too high: it also includes bank account holdings / revenues.
I assume B-securities are off in case of a non-CH bank. In case it would be a CH bank, cash accounts would account for A, which would cause the same issues for A-securities.

I assume the root issue is in `total.py`, where `svGrossRevenueA` / `svGrossRevenueB` /  `svTaxValueA` / `svTaxValueB` contain the total of all A / B values (both cash and securities). Those things should stay separate so that the correct subtotals can be used in r`ender.py`. For the summary, those values can be summed up.

vroonhof:
Well spotted...

PR #199, opened 2026-02-18
Fix Issue #197: Prevent TotalCalculator from overwriting spec fields

vroonhof:
Fixes Issue #197 by ensuring that the `TotalCalculator` writes calculated totals to internal fields (e.g., `internalTotalTaxValue`, `internalTotalGrossRevenueA`) rather than overwriting the standard eCH-0196 export fields (`totalTaxValue`, etc.).

Changes:
- Added internal fields to `TaxStatement`, `ListOfSecurities`, `ListOfBankAccounts`, `ListOfLiabilities`, `BankAccount`, and `LiabilityAccount` models in `src/opensteuerauszug/model/ech0196.py`. These fields are marked `exclude=True` to prevent serialization.
- Modified `TotalCalculator` in `src/opensteuerauszug/calculate/total.py` to populate these internal fields during `FILL` and `OVERWRITE` modes. In `VERIFY` mode, it verifies the calculated values against the existing spec fields.
- Updated `src/opensteuerauszug/render/render.py` to use a new helper `get_val_with_fallback` that prioritizes internal fields for display, falling back to spec fields if internal fields are missing (e.g., if calculator wasn't run). This ensures the PDF report shows the calculated values while preserving the original imported values in the data model for export.
- Updated unit tests in `tests/calculate/test_total.py` to verify calculations against internal fields.

This ensures that imported tax statements preserve their original totals in the XML export, while providing accurate recalculated totals for the generated PDF report.

---
*PR created automatically by Jules for task [14800570527800998343](https://jules.google.com/task/14800570527800998343) started by @vroonhof*

### DIFF
--- a/src/opensteuerauszug/calculate/total.py
+++ b/src/opensteuerauszug/calculate/total.py
@@ -72,7 +72,28 @@ class TotalCalculator(BaseCalculator):
         # that is OK as rounding is idempotent.
         rounded_value = round_accounting(value)
         self._set_field_value(model, field_name, rounded_value, path)
-    
+
+    def _update_field(self, model: Any, spec_field: str, internal_field: str, value: Decimal, path: str) -> None:
+        """
+        Update a field, handling the distinction between spec fields (verification only)
+        and internal fields (setting values).
+
+        Args:
+            model: The model to update
+            spec_field: The name of the field in the eCH-0196 spec (for verification)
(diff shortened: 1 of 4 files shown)


vroonhof:
This is garbage.. flash jules misunderstood solution constraint for main instruction.

PR #205, opened 2026-02-20
PDF: Add total number of pages to footer

pet-zh:


vroonhof:
nice that is so cleab.

PR #209, opened 2026-02-21
fix(ibkr): handle WHTAX cash transactions not linked to a security

MariusLutz:
## Summary

- IBKR's Stock Yield Enhancement Program generates withholding tax entries (`CashAction.WHTAX`) that are not associated with a specific security (no `conid`). These fell through to the `else` branch and raised a `ValueError`.
- This PR adds `CashAction.WHTAX` to the non-security cash transaction handling, treating it as a tax-relevant bank payment -- consistent with how `BROKERINTRCVD` is already handled.

## Context

When processing a real IBKR Flex Query that included yield enhancement income with 30% withholding, the importer crashed with:

```
ValueError: CashTransaction type CashAction.WHTAX is not supported for WITHHOLDING @ 30% ON YIELD ENHANCEMENT INCOME FOR FEB-2025
```

### DIFF
--- a/tests/importers/ibkr/test_ibkr_importer.py
+++ b/tests/importers/ibkr/test_ibkr_importer.py
@@ -2008,6 +2008,11 @@ SAMPLE_IBKR_FLEX_XML_WHTAX_NO_SECURITY = """
       </CashTransactions>
       <CashReport>
         <CashReportCurrency accountId="U1234567" currency="CHF" endingCash="1000.00" />
+      </CashReport>
+    </FlexStatement>
+  </FlexStatements>
+</FlexQueryResponse>
+"""
 
 def test_withholding_tax_without_security_creates_bank_payment(sample_ibkr_settings):
     """Withholding tax not linked to a security (e.g. yield enhancement) should


pet-zh:
@MariusLutz Thanks for you PRs. When fixing importer related things, please always post relevant full lines from flex export (anonymized).

If you re using an agent, generating a test should be very easy since we have many such samples. Otherwise we can also generate one follow up.

MariusLutz:
I used an agent to add a test, as suggested. Please let me know if this is helpful or further refinements are necessary.

vroonhof:
hi @MariusLutz thank you for adding the test. it is very helpful

Do you get any indication (e.g. from the normal statements) which jurisdication the withholding was for? Swiss withholding tax is represented differently on these entries. 
Any indiction if it is always like in your example where the withholding tax if fully cancelled again [so the end effect on tax income is zero]

MariusLutz:
Let me preface I am not an expert and this is merely my interpretation based on my experience.
1. In my excerpts no information on issuerCountryCode or isin is given. 
2. IBKR's Stock Yield Enhancement Program (SYEP) lends shares for short periods; the income is classified as US-source by IBKR, which is why they initially withhold 30%.
3. The withhold 30% are canceled in my dataset every single time because of the W-8BEN and double taxation treaty. I would thus assume it's not guaranteed to always be fully cancelled but regular procedure. In case processing lags or the cancellation does not take place it would show up as an unmatched income line.

vroonhof:
I also have stock yield program but it is for swiss shares etc and i don't
see this. So perhaps it is indeed for income from lending us securities in
your case.
Anyway seems ok at this point.

It is just particularly dangerous to have negative entries in the income
column

vroonhof:
Need to be at my computer to resolve the merge

PR #210, opened 2026-02-21
fix(calculate): downgrade stock split validation errors to warnings

MariusLutz:
## Summary

- Downgrades stock split validation errors to warnings so that processing can continue when the broker data doesn't contain the exact corporate action mutations expected by the Kursliste.
- Affects both same-ISIN splits (e.g. reverse splits) and cross-ISIN splits (where the valor number changes).
- The user is prompted to verify the affected securities manually in the generated output.

## Context

When processing real IBKR data against the 2025 Kursliste, the calculator crashed on two securities:

1. **NL0015436031** -- a reverse stock split on 2025-10-22 that IBKR did not record as a corporate action in the Flex Query export.
2. **LU1781541179** -- a cross-ISIN split on 2025-02-21 (new valor 131443426) where IBKR's corporate action data didn't match the Kursliste expectation.

These are legitimate discrepancies between the Kursliste and broker data. A hard error prevents the entire Steuerauszug from being generated, which is worse than producing it with a warning for the user to check.

## Changes

All 6 `raise ValueError(...)` calls in `_validate_stock_split` are replaced with `logger.warning(...)` + `return`, covering:
- Missing same-ISIN split mutations
- Same-ISIN split ratio mismatches
- Missing cross-ISIN removal mutations
- Missing new security for cross-ISIN splits
- Missing addition mutations on the new security
- Cross-ISIN addition quantity mismatches

### DIFF
--- a/src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
+++ b/src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
@@ -83,11 +83,13 @@ class KurslisteTaxValueCalculator(MinimalTaxValueCalculator):
         self.flag_override_provider = flag_override_provider
         self._current_kursliste_security = None
         self._missing_kursliste_entries = []
+        self._stock_split_warnings: List[dict] = []
         self._previous_year_exdate_warnings = []
         self._all_securities: List[Security] = []
 
     def calculate(self, tax_statement):
         self._missing_kursliste_entries = []
+        self._stock_split_warnings = []
         self._previous_year_exdate_warnings = []
         # Collect all securities across all depots so that cross-security
         # split validation (valorNumberNew) can look up the target security.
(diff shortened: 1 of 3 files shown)


vroonhof:
Thank your contributions.  I am a little bit hesitant to downgrade the errors by default especially as thing are still very chatty right now and still easy to miss. Perhaps this can be an option or a CriticalWarning (e.g. as visible in the PDF) instead.

I am also wondering ab bit how IBKR does represent these actions without being incorrect in the balances. for me at least in case of corporate actions those are the only part of the export where the mutations to the amount of stock are represented (this is also what makes me more hesitant).

MariusLutz:
Thanks for the feedback – you raise good points.

I agree that a plain logger.warning is too easy to miss. Using CriticalWarning (rendered on the PDF instructions page) is a much better fit here – it's the same pattern used for MISSING_KURSLISTE and UNMAPPED_SYMBOL, and the issue is exactly in that severity class. I'll rework the PR to emit a CriticalWarning instead of just logging, and keep the processing going rather than aborting. A new category like STOCK_SPLIT_MISMATCH would make it filterable too.

In my case, the Kursliste had a split entry for a security where IBKR used a slightly different effective date or the action hadn't propagated into my export period yet. The entire statement generation failed, even though the position balances were otherwise correct. A CriticalWarning on the PDF would flag it clearly without blocking the rest of the output.

PR #212, opened 2026-02-21
XML: Fix schema validation error (missing attributes)

### LINKED ISSUE #208: Validator error: Attribute 'totalGrossRevenueIUP' must appear on element 'listOfSecurities'.
Looks like this issue got introduced recently, same import data worked until a few days ago if I remember correctly. I have not checked yet which change caused it.

This also breaks importing into ZH tax software.

pet-zh:
totalGrossRevenueIUP and totalGrossRevenueConversion got lost in 18ea999

Resolves #208


vroonhof (approved):
Thanks

PR #213, opened 2026-02-21
chore: Add XML schema validation to integration test

pet-zh:
Related to #212

### DIFF
--- a/tests/test_integration_ibkr_2025.py
+++ b/tests/test_integration_ibkr_2025.py
@@ -2,8 +2,23 @@ import pytest
 from typer.testing import CliRunner
 from pathlib import Path
 from opensteuerauszug.steuerauszug import app
+import lxml.etree as ET
 
 runner = CliRunner()
+
+class _LocalXsdResolver(ET.Resolver):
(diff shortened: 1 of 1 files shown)


pet-zh:
@vroonhof It is expected that this fails because main is currently broken. I will rebase this change once #212 is merged.

BTW, would be nice if you could enabled automated [copilot PR reviews by default](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/configure-automatic-review#configuring-automatic-code-review-for-a-single-repository), then I would get instant feedback.

vroonhof:
> @vroonhof It is expected that this fails because main is currently broken. I will rebase this change once #212 is merged.
> 
> BTW, would be nice if you could enabled automated [copilot PR reviews by default](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/configure-automatic-review#configuring-automatic-code-review-for-a-single-repository), then I would get instant feedback.

Copilot review is on. The setting is exactly as in the docs. Not sure why it is not triggering for you. Perhaps it is unclear who's plan pays for it..

Issue #148, opened 2026-02-08
PDF: Attach (embed) original import data file


pet-zh:
For audit reasons, and also providing an (official) proof, we can add the unaltered import data file (e.g. flex query) as PDF attachment. In case such an attachment is present, it should clearly be described on the overview or description page.

A library such as [tikepdf](https://pikepdf.readthedocs.io/en/latest/topics/attachments.html) should be able to do this.

vroonhof:
The standard actually allows attaching things on the xml. But they can only
be pdf documents

pet-zh:
Hm, the link to the library shows how to attach a .md file to the PDF at least. Not sure if thos follows PDF standards, but I also think I have seen PDF which attach files other than PDFs. Do you have a reference which says this is not allowed in the PDF standard?

martinkretzschmar:
When you upload a PDF to the Zurich tax web app, it acts a bit as if the pdf gets converted to bitmaps, or at least reduced to a (sub)set of pages extracted from the PDF. PDF-specific features like embedded files might not survive. Concatenating the reportlab-generated pdf and the original PDF (eg. with the linked pikepdf) might be more practical than attaching.

pet-zh:
@martinkretzschmar I have also noticed this. I hope this rendering process only happens for display reasons, and the original file still gets transferred in full quality. But you never know.

We could also add those 3 options:
- embed ibflex XML
- embed IBKR PDF unaltered into PDF
- attach PDF to tax statement PDF visible at the end (need to check in spec if this is allowed at all, probably it also needs page barcodes and such and needs to be before the barcodes pages)

vroonhof:
At the very end of the tax return process in Zürich you can download the
return. This will download a zip file. This file seems to contain the
original pdfs(or something very close) renamed to standardized names based
on the statement ID. So the data at least gets preserved somewhere

pet-zh:
FYI, pre-pending or appending PDFs is now supported. The idea of attachichign the original importer data (e.g. IB flex XML) is on hold, see here: https://github.com/vroonhof/opensteuerauszug/pull/214#issuecomment-3977768934.

Anybody is welcome to comment on this idea.

PR #214, opened 2026-02-21
PDF: Embed original import data in PDF

pet-zh:
Part of #148

### DIFF
--- a/src/opensteuerauszug/render/render.py
+++ b/src/opensteuerauszug/render/render.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Dict, Any, Optional, Union
 from decimal import Decimal, ROUND_HALF_UP
 import zlib
+import zipfile
 from PIL import Image as PILImage
 import html
 
@@ -1893,6 +1894,7 @@ def render_tax_statement(
     output_path: Union[str, Path],
     override_org_nr: Optional[str] = None,
     minimal_frontpage_placeholder: bool = False,
+    embed_import_data: bool = True,
 ) -> Path:
     """Render a tax statement to PDF.
     
@@ -1902,7 +1904,9 @@ def render_tax_statement(
         override_org_nr: Optional override for organization number (5 digits)
         minimal_frontpage_placeholder: If True, replace the summary on the first
             page with a placeholder suitable for minimal tax statements
-        
+        embed_import_data: If True, embed import data (if available) zipped as
+            PDF attachment
+
     Returns:
         Path to the generated PDF file
     """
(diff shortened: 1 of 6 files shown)


src/opensteuerauszug/render/render.py
```diff
@@ -1940,10 +1944,65 @@ def _canvas_maker(*args, **kwargs):
...
+            logger.debug(f"Could not compress page content streams: {e}. Continuing without compression for this page.")
+
+    if embed_import_data and tax_statement.import_data_path:
+        try:
+            # Create a zip file containing import files
+            zip_buffer = io.BytesIO()
+            with zipfile.ZipFile(zip_buffer, 'w', zipfile.ZIP_DEFLATED) as zip_file:
+                import_path = Path(tax_statement.import_data_path)
+                files_to_add = [import_path] if import_path.is_file() else list(import_path.iterdir()) if import_path.is_dir() else []
+                allowed_extensions = {'.xml', '.json', '.csv', '.pdf'}
```
pet-zh, line 2175: Please double check if that includes all relevant files (also for schwab for example).

vroonhof:
This is one of the changes where you would love have access to the tax office technical contacts to see uf they would be able to read it.

It will take some time to review this sorry.

pet-zh:
Yes, would be nice. I have checked the spec, and I have not found anything related to attachments. It mentions that the ECH XML could be embedded, but would need to be signed with a valid cert. For that reason I decided to pack non-ECH things into a ZIP so it should not conflict with XML files (some tax software could think the IB Flex XML is a ECH XML and fail importing).

I have tested this change against ZH online tax software, and it was able to process this.

If some tax software fails to read it, the user can just re-generate the PDF using the config flag. So it should not be a blocker. And once the tax software is able to import, I do not think tax office will have issues processing it.

pet-zh:
FYI, I have a follow-up change (would create a lot of merge conflicts otherwise) which allows to prepend an original bank or broker PDF.  It looks like only prepending to the PDF is allowed, otherwise a page barcode would be needed (I have some real world examples which also prepend pages without barcodes on pages prepended pages, but all which are appending pages after the summary page have barcodes on the pages included).

I have tested importing the whole bundle (prepended IB activity statement PDF and embedded IB flex report XML zipped) into my real tax tool - worked without any issues.

vroonhof:
sorry @pet-zh  for not having time to push on this (and I independently invented the preoebd-pdf feature). How strongly do you feel we need the embedding of the data if we can have PDFs pre-pended as I am not sure the tax office workflow allows extracting zips and how useful it is for them to see a bunch XML, json and CSV.

architecture wise I think the right way if the importer reported back the files it actually used, that allows the schwab one to avoid any PDFs for other files it skipped in the directly.

Once they are generated I intend to post-pend the 1042-S documents from and IBKR and Schwab, pre-pend an explanation letter and see if that works with my real tax return and if the tax office really cares about the original statment. (perhaps I should try post-pending a 100 page doc and se if tax-statement.jar really cares).

pet-zh:
I do not have a strong opinion here. Tax offices most likely prefer an easy-to-read PDF attachment over a zipped XML file.

From a user’s point of view, it would be easier to use this option since it would not require downloading another report from the broker.

For this year, I will most likely prepend the IB activity report to the PDF and see how it goes (in the past, I only submitted the activity report after manually filling in the positions) for DA-1.

I am fine with putting this on hold for now, and if we receive more user feedback or stronger reasons to include the original data, we can move forward.

vroonhof:
Let's park this.

PR #216, opened 2026-02-21
Automate kursliste download

seirl:
Automated the download of Kursliste XML files from the ICTax website. Added a new CLI module opensteuerauszug.kursliste and implemented the download, unzip, and storage logic. Verified with unit tests, and manual testing:

```
python -m opensteuerauszug.kursliste download --year 2025
```

As next steps after this pull request, I think we could automate the process even further by:

* integrating `scripts/convert_kursliste_to_sqlite.py` to the new `opensteuerauszug.kursliste` module (and make the download command also do this by default)
* reading and writing these files in `$XDG_DATA_HOME/opensteuerauszug/` so that you can run this script in any directory without needing the `data/` folder.

### DIFF
--- a/data/kursliste/kursliste.md
+++ b/data/kursliste/kursliste.md
@@ -62,7 +62,7 @@ You can automatically download the latest Kursliste for a given year using:
 python -m opensteuerauszug.kursliste download --year 2024
 ```
 
-This will fetch the official XML file from the ESTV API and place it in this directory as `kursliste_2024.xml`.
+This will fetch the official XML file from the ESTV website using its internal API and place it in this directory as `kursliste_2024.xml`.
 
 ### Manual Data Placement
 


src/opensteuerauszug/kursliste/downloader.py
```diff
@@ -0,0 +1,140 @@
...
+logger = logging.getLogger(__name__)
+
+API_URL = "https://www.ictax.admin.ch/extern/api/xml/xmls.json"
+SESSION_URL = "https://www.ictax.admin.ch/extern/api/authentication/session.json"
+HOME_URL = "https://www.ictax.admin.ch/extern/en.html"
+DOWNLOAD_BASE_URL = "https://www.ictax.admin.ch/extern/api/download"
+
+# Standard headers to avoid 403
+HEADERS = {
+    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36",
```
vroonhof: is the IC tax website really that big into antiscraping that we have to pretend to be a real browsr as well?

src/opensteuerauszug/kursliste/downloader.py
```diff
@@ -0,0 +1,140 @@
...
+    "Accept": "application/json, text/plain, */*",
+    "Accept-Language": "en-US,en;q=0.9",
+    "Content-Type": "application/json",
+    "Origin": "https://www.ictax.admin.ch",
+    "Referer": HOME_URL
```
vroonhof: same here... are they enforcing a referrer we have to fake?

src/opensteuerauszug/kursliste/downloader.py
```diff
@@ -0,0 +1,140 @@
...
+}
+
+def download_kursliste(year: int, destination_dir: Path) -> Path:
+    """
+    Downloads the Kursliste for the given year from ICTax API.
```
vroonhof: this isn't the official API (I think that is open to allow listed providers only). The official API creates a kursliste just for your instutition on demand (which is apparently expensive).

vroonhof (commented):
Nice. Can you retitle this to not pretend it is the official API. we are just simulating the browser here.

src/opensteuerauszug/kursliste/downloader.py
seirl: Not needed, fixed :-)

seirl:
@vroonhof PTAL :-)

Issue #218, opened 2026-02-21
IBKR importer crashes on short options positions (covered calls)


ksfp941:
## Description

The IBKR importer fails when the Flex Query export contains short options positions (e.g. written covered calls). The tool errors out during the import phase with a "Negative balance" error because it does not expect positions with negative quantities.

## Error

```
ValueError: Negative balance computed for security 200001 (start 0, end -12)
```

This is raised in `ibkr_importer.py` at the balance validation check:
```python
if opening_balance < 0 or closing_balance < 0:
    raise ValueError(...)
```

Short (written) options naturally have negative quantities — e.g. selling 12 covered call contracts results in a year-end position of -12. The tool currently treats any negative balance as invalid.

## Minimal reproduction file

The XML below is a fully anonymized, minimal IBKR Flex Query export that reproduces the issue. It contains:
- **1 stock** (XXXX): 1,200 shares bought and held at year-end
- **1 ETF** (YYYY): 50 shares bought and held at year-end
- **12 short call contracts** on XXXX: sold in two batches (5 + 7), held short at year-end — this triggers the error
- Dividend, withholding tax, broker interest, and cash report entries

The stock position of 1,200 shares covers the 12 short contracts (100 shares per contract), making this a standard covered call setup.

[minimal_short_options_repro.xml](https://github.com/user-attachments/files/25461015/minimal_short_options_repro.xml)

<details>
<summary>minimal_short_options_repro.xml (click to expand)</summary>

```xml
<FlexQueryResponse queryName="Annual Tax Report" type="AF">
<FlexStatements count="1">
<FlexStatement accountId="U9999999" fromDate="2025-01-01" toDate="2025-12-31" period="" whenGenerated="2026-02-01;10:00:00">
[...]

PR #224, opened 2026-02-22
Omit `security.taxValue` in case of 0 end-of-period position

pet-zh:
From the specs, p.32:
> Nur wenn der Endbestand eines Titels zum Periodenende (periodTo) grösser als 0 ist, dann ist der Steuerwert anzugeben.

### DIFF
--- a/src/opensteuerauszug/calculate/cleanup.py
+++ b/src/opensteuerauszug/calculate/cleanup.py
@@ -428,13 +428,15 @@ class CleanupCalculator:
                                     candidate = security.stock[find_index]
                                     if candidate.referenceDate == period_end_plus_one and not candidate.mutation:
                                         # First balance after the period end is the end balance of the period
-                                        security.taxValue = SecurityTaxValue(
-                                            referenceDate=self.period_to,
-                                            quotationType=candidate.quotationType,
-                                            quantity=candidate.quantity,
-                                            balanceCurrency=candidate.balanceCurrency,
-                                            balance=candidate.balance,
-                                            unitPrice=candidate.unitPrice)
+                                        # Only create taxValue if closing balance is not zero
+                                        if candidate.quantity is not None and candidate.quantity != Decimal('0'):
+                                            security.taxValue = SecurityTaxValue(
+                                                referenceDate=self.period_to,
+                                                quotationType=candidate.quotationType,
+                                                quantity=candidate.quantity,
+                                                balanceCurrency=candidate.balanceCurrency,
+                                                balance=candidate.balance,
+                                                unitPrice=candidate.unitPrice)
 
                             # TODO Should we ensure the balances at the start and end of the period are
                             #       present here instead of in the importers?.
--- a/src/opensteuerauszug/importers/schwab/schwab_importer.py
+++ b/src/opensteuerauszug/importers/schwab/schwab_importer.py
@@ -553,12 +553,14 @@ def convert_cash_positions_to_list_of_bank_accounts(
                     break
         
         if closing_stock_entry:
-            bank_account.taxValue = BankAccountTaxValue(
-                referenceDate=period_to, # Tax value is as of end of period_to
-                name="Closing Balance",
-                balanceCurrency=closing_stock_entry.balanceCurrency,
-                balance=closing_stock_entry.quantity # Quantity of cash is its balance
-            )
+            # Only create taxValue if closing balance is not zero
+            if closing_stock_entry.quantity is not None and closing_stock_entry.quantity != Decimal('0'):
+                bank_account.taxValue = BankAccountTaxValue(
+                    referenceDate=period_to, # Tax value is as of end of period_to
+                    name="Closing Balance",
+                    balanceCurrency=closing_stock_entry.balanceCurrency,
+                    balance=closing_stock_entry.quantity # Quantity of cash is its balance
+                )
         accounts.append(bank_account)
     return ListOfBankAccounts(bankAccount=accounts)
 
--- a/tests/calculate/test_cleanup.py
+++ b/tests/calculate/test_cleanup.py
@@ -1842,3 +1842,191 @@ class TestCleanupCalculatorIDGeneration:
         calculator = CleanupCalculator(period_from=DEFAULT_TEST_PERIOD_FROM, period_to=DEFAULT_TEST_PERIOD_TO, importer_name="STRIPCASE", enable_filtering=False)
         calculator.calculate(statement)
         assert statement.id == expected_id
+
+
+class TestCleanupCalculatorClosingBalanceQuantity:
+    """Tests for SecurityTaxValue creation based on closing balance quantity."""
+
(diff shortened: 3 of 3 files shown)


src/opensteuerauszug/calculate/cleanup.py
```diff
@@ -366,13 +366,15 @@ def calculate(self, statement: TaxStatement) -> TaxStatement:
...
                                     if candidate.referenceDate == period_end_plus_one and not candidate.mutation:
                                         # First balance after the period end is the end balance of the period
-                                        security.taxValue = SecurityTaxValue(
-                                            referenceDate=self.period_to,
-                                            quotationType=candidate.quotationType,
-                                            quantity=candidate.quantity,
-                                            balanceCurrency=candidate.balanceCurrency,
-                                            balance=candidate.balance,
-                                            unitPrice=candidate.unitPrice)
+                                        # Only create taxValue if closing balance is not zero
```
vroonhof, line 431: I think is wrong. The spec says

Es ist nur dann eine 0 beim Steuerwert anzugeben, wenn der Steuerwert tatsächlich 0 ist. 

which mean that if we know balance is zero, it should be zero. In my view that makes sense

src/opensteuerauszug/importers/schwab/schwab_importer.py
```diff
@@ -553,12 +553,14 @@ def convert_cash_positions_to_list_of_bank_accounts(
                     break
         
         if closing_stock_entry:
-            bank_account.taxValue = BankAccountTaxValue(
-                referenceDate=period_to, # Tax value is as of end of period_to
-                name="Closing Balance",
-                balanceCurrency=closing_stock_entry.balanceCurrency,
-                balance=closing_stock_entry.quantity # Quantity of cash is its balance
-            )
+            # Only create taxValue if closing balance is not zero
```
vroonhof, line 556: same as above. this just smells wrong. it must be possibe to assert positively that the tax value is zero.

src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
```diff
@@ -203,6 +203,8 @@ def _handle_SecurityTaxValue(self, sec_tax_value: SecurityTaxValue, path_prefix:
                     self._set_field_value(sec_tax_value, "balanceCurrency", "CHF", path_prefix)
                     self._set_field_value(sec_tax_value, "kursliste", True, path_prefix)
                     return
+        else:
```
vroonhof: note that the moment this case leads to an unsupported setup, but I agree it is correct to set the flag.
we need to find a way to clear it again in the fill in taxvalue calculator where w do it based on broker that

src/opensteuerauszug/calculate/cleanup.py
vroonhof, line 431: repeating for review (turns out I don't know how drive github). Pleas don't do this. I think we should zero if that is really the tax value. if we end up with zero due to upstream issue that we don't support we should handle that some other way.

pet-zh: From the specs, p.32:
>  Nur wenn der Endbestand eines Titels zum Periodenende (periodTo) grösser als 0 ist, dann ist der
Steuerwert anzugeben. 

I have also seen omitting the whole `taxValue` element in DL examples. And verified using the official verifier tool.

I think your excerpt is referring to the case when a `taxValue` element is present.

Can you probably double check this change against your local test set of real world statements?

Just to also give some context why I have discovered this case: In #218, there are a lot of positions with 0 end of year balance, and none of those have a kursliste entry. That caused many warnings due to missing prices. With this optimizations, we do not even trigger a kursliste lookup anymore in case the price is irrelevant anyway.

src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
pet-zh: Having a `taxValue` element without a `unitPrice` and `undefined` not set to `1`, causes a validation error in the official verifier tool. So we should set it in any case in order to produce a valid statement.

In case there is a place where we would set `unitPrice` later on, we of course have to clear out undefined again. Do you know such a place?

specs, p.35:
> Das Flag undefiniert für einen nicht oder noch nicht verfügbaren Steuerwert. Falls das Flag zutrifft, so muss der fixe Werte 1 (true) gesetzt werden.

pet-zh:
@vroonhof I think I have addressed all of your concern. Can you please take another look here?

vroonhof:
> @vroonhof I think I have addressed all of your concern. Can you please take another look here?

I think you disagreed with me, but I  am not convinced. I still think a zero tax value of is valid positive assertion. I am traveling so don't have time to look at my other examples.

if you want we can just do the setting of the undefined flag (and then override it in the fill_in calculator if we are making a guess). in fact in "minimal" mode we should probably always set the undefined flag?

src/opensteuerauszug/calculate/cleanup.py
pet-zh: > Es ist nur dann eine 0 beim Steuerwert anzugeben, wenn der Steuerwert tatsächlich 0 ist.

Note the term "Steuerwert": Steuerwert is `unitPrice` in this case in my understanding (also if you check the term ["Steuerwert" on ICtax](https://www.ictax.admin.ch/extern/de.html#/security/4354003/20251231:~:text=Valoren%2DNr.-,Steuerwert,-Ertrag%20/%20Rendement%20/%20Reddito)). And this is not 0, even when position balance is 0.

vroonhof:
I finally had time to double check and at least one real world example I have (from TrueWealth) has taxValue with balance zero

```
        <taxValue referenceDate="2024-12-31" quotationType="PIECE" quantity="0" balanceCurrency="CHF" unitPrice="117.9" value="0"/>
```
This is likely also why I didn't run into this before (because most of my investments are buy and hold , just the roboadvisors or stock mergers produce end of year zero balances). The TrueWealth one deviates from the spec and "the usual way" in many places, so it is not the best one as guide.

vroonhof:
so I think  your right that ommiting it would be more spec compliant. some things to watch
- Spec actually says "grösser als 0" (which may mean something for the short handling if it ever comes)
- I don't like the side efffect of no longer checking for the kursliste entry. For that we would also have to assert there are no tax relevant events throughout the year which is hard to know without the kurstliste.

vroonhof:
actually reading the code it looks we will still warn as long as we have seen the security line..

src/opensteuerauszug/importers/schwab/schwab_importer.py
vroonhof: I am going to revert this part. lets handle this by clearing in cleanup (which makes the logic indendent of imprter)

vroonhof:
> actually reading the code it looks we will still warn as long as we have seen the security line..

this turns out not be true due to accidental effect. we don't real the year from the global state but instead from the tax value and bail out.

vroonhof:
Closing in favour of a continuation branch with additional improvements — see #276

PR #226, opened 2026-02-22
IBKR: Do not set trade price if 0

pet-zh:
Official validation tool does not allow 0 value unit price (failing check: SECURITY-04-11 - Stückpreis angegeben). Since this is an optional attribute, omit it.

Such 0 values trades can be found in the example XML of issue #218.

### DIFF
--- a/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
+++ b/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
@@ -417,7 +417,7 @@ class IbkrImporter:
                         referenceDate=trade_date,
                         mutation=True,
                         quantity=quantity,
-                        unitPrice=trade_price,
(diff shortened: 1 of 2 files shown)


vroonhof:
I wonder if we should start adding knowledge like this to our brand new validation phae.

pet-zh:
> I wonder if we should start adding knowledge like this to our brand new validation phae.

At this point, I'm not sure how we should proceed with validation, particularly when it comes to XSD schema validation. The official tool contains many test cases, but they are not documented. By decompiling the tool, it's possible to understand what it attempts to do in most cases. We could ask an agent to document all the test cases from the official tool and reimplement them, but I’m not sure whether this might trigger potential copyright issues, especially since the official JAR doesn’t contain any license information. For now, I would probably just run the official validator tool periodically or even integrate it into the GH Actions workflow.

PR #227, opened 2026-02-22
Convert negative bank account holdings into liabilites

pet-zh:
Remove `LiabilityAccount.liabilityCategory` since it is not a valid attribute according to XSD.

### DIFF
--- a/tests/calculate/test_cleanup.py
+++ b/tests/calculate/test_cleanup.py
@@ -1236,7 +1236,7 @@ class TestNegativeBankAccountBalance:
         assert liability.taxValue.balance == Decimal("1500.50")
         assert liability.totalTaxValue == Decimal("1500.50")
         assert liability.bankAccountNumber == bank_account.bankAccountNumber
-        assert liability.bankAccountName == bank_account.bankAccountName
+        assert bank_account.bankAccountName in liability.bankAccountName
         assert liability.bankAccountCountry == bank_account.bankAccountCountry
         assert liability.bankAccountCurrency == bank_account.bankAccountCurrency
 


src/opensteuerauszug/render/render.py
```diff
@@ -621,77 +621,164 @@ def create_summary_table(data, styles, usable_width):
     return KeepTogether([summary_table, Spacer(1, 2*mm)])
 
 # --- Liabilities Table Function ---
-def create_liabilities_table(data, styles, usable_width):
-    """Creates a table displaying liabilities information.
-    
+def create_liabilities_table(tax_statement, styles, usable_width):
```
vroonhof, line 624: for follow up. If we are going to create liabilities then we should probably also add them to the overview page. Unfortunately in the standard layout that is put at the bottom (left) where we put our notices.

I am including this because it is an improvement we should definately not create negative bank account values.

src/opensteuerauszug/calculate/cleanup.py
```diff
@@ -252,6 +258,45 @@ def calculate(self, statement: TaxStatement) -> TaxStatement:
...
+                        logger.info(
+                            f"  BankAccount {account_id}: Found negative balance {bank_account.taxValue.balance}. "
+                            f"Creating liability account and setting balance to 0."
+                        )
+
+                        # Create liability account from the negative bank account
+                        liability_account = LiabilityAccount(
+                            iban=bank_account.iban,
+                            bankAccountNumber=bank_account.bankAccountNumber,
+                            bankAccountName=bank_account.bankAccountName or BankAccountName(account_id),
```
vroonhof: I feel we should add some reference in name to say this is a negative balance. I agree it is part of the context but I think it would also be clearer for the user.

src/opensteuerauszug/calculate/cleanup.py
```diff
@@ -276,6 +321,17 @@ def calculate(self, statement: TaxStatement) -> TaxStatement:
         else:
             logger.info("No bank accounts found to process.")
 
+        # Add any liabilities created from negative bank account balances
+        if liabilities_to_add:
```
vroonhof, line 330: not 100% clear why we are collecting these in a side array first instead of just directly adding them to the the tax statement,

PR #235, opened 2026-02-24
Convert negative interest payments into liabilities payments

pet-zh:
Negative bank account payments are invalid according to the official validator and must therefore be recorded as liabilities.

### DIFF
--- a/src/opensteuerauszug/calculate/cleanup.py
+++ b/src/opensteuerauszug/calculate/cleanup.py
@@ -8,7 +8,7 @@ import logging
 from opensteuerauszug.model.ech0196 import (
     SecurityTaxValue, TaxStatement, SecurityStock, BankAccountPayment, SecurityPayment,
     Client, ClientNumber, CantonAbbreviation, LiabilityAccount, LiabilityAccountTaxValue,
-    ListOfLiabilities, BankAccountName, CountryIdISO2Type, CurrencyId
+    ListOfLiabilities, BankAccountName, CountryIdISO2Type, CurrencyId, LiabilityAccountPayment
 )
 from opensteuerauszug.model.critical_warning import CriticalWarning, CriticalWarningCategory
 from opensteuerauszug.util.sorting import find_index_of_date, sort_security_stocks, sort_payments, sort_security_payments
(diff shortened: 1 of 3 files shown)


src/opensteuerauszug/calculate/cleanup.py
```diff
@@ -323,6 +320,74 @@ def calculate(self, statement: TaxStatement) -> TaxStatement:
                         else:
                             logger.info(f"  BankAccount {account_id}: Payment filtering skipped (tax period not fully defined).")
                     # No log if filtering is disabled globally
+
+                    # Move negative payments to liabilities
```
vroonhof, line 325: this is not actually true for all negative payments.. 

For example for 
interest on  cash loans [and overdrafts] --> liability payments
interest on short --> probably liability payments because above [but needs double checkimg]
negative interest [e.g. CHF] --> **cost of trading.**

(it is one reason Swis banks don't actually report negative interest, but charge "Guthabengebühr")

One reason we have been dropping negative IBKR amoutns on the floor is because I am not 100% sure which of the above are these when we see "Broker Interested Paid"

pet-zh: > negative interest [e.g. CHF] --> cost of trading.

Good catch! This is a valid point. We need to understand how IB reports this type of interest - will look into this.

pet-zh:
@vroonhof I have removed the importer related part for now, I will investigate how IBKR handles this and create a follow-up PR.

The other parts of this PR I would propose to get landed in order to reduce the chances of merge conflics:
- adjusting the renderer allows rendering e.g. DL example 1:1 in interest related parts
- negative  bank account payments are invalid as per official validator in any case - they need to be converted into liabilities

pet-zh:
@vroonhof I was able to clarify the different types of IBKR interest payments based on CHF and USD examples from [vtandchill2025.xml](https://github.com/vroonhof/opensteuerauszug/blob/main/tests/samples/import/ibkr/vtandchill_2025.xml) and [eTax_report_anonymised_2025.xml](https://github.com/vroonhof/opensteuerauszug/blob/main/tests/samples/import/ibkr/eTax_report_anonymised_2025.xml). See updated description above for details.

pet-zh:
I have now extracted the IBKR importer related stuff into a separate PR: https://github.com/vroonhof/opensteuerauszug/pull/249

This precursor PR only makes sure that no invalid XML is created in case importer imports such payments, and also values are rendered.

PR #243, opened 2026-02-28
Add uv.lock, for reproducibility and packaging with uv2nix

seirl:
This is important for this specific project because it uses vendoring and forks. As soon as the forks are merged into the main packages, this could potentially be removed. In the meantime, if we want to package this software for NixOS, we can't use the existing packages, so we need to maintain an up-to-date uv.lock with the pinned dependencies.


vroonhof:
Is there a good way to keep this up to date while not mandating the use of uv to iterate or maintain this?

vroonhof:
e.g. what if pyproject and uv.lock get out of sync and/or I rev vroonhof/ibflex?

seirl:
There seems to be a GitHub action for this: https://github.com/hbelmiro/uv-lock-check

Should I add it?

seirl:
@vroonhof OK, I added it to the workflow. If you forget to add a dependency, it gives you this error in the CI now:

image

vroonhof:
would this work if we provided pylock.toml instead. 

(sorry for being so petty, but one of the environments where I develop (used to) make it hard to install UV (because it is not pure python). this has since been fixed [ and things like github code spaces exist] but it would still be nice to not make it hard requirement.

seirl:
It looks like there's no easy way to package a pylock.toml to a Nix environment as far as I can see.
(Another option would be to wait until your forks & PRs are merged in the mainline projects, and this would also allow us to put it directly in nixpkgs, but this is obviously less easy to get done.)

vroonhof:
PR418 will finally commit to locking with uv.lock. Thanks for starting this.

PR #244, opened 2026-02-28
chore: Fix typos

pet-zh:


src/opensteuerauszug/model/kursliste.py
```diff
@@ -710,7 +710,7 @@ class Kursliste(PydanticXmlModel, tag="kursliste", nsmap=NSMAP):
     # --- Attributes aligned with XSD ---
     version: Annotated[str, StringConstraints(pattern=r"2\.[02]\.0\.\d")] = attr()
     creationDate: datetime.datetime = attr()
-    referingToDate: Optional[datetime.date] = attr(default=None)
```
vroonhof, line 713: this is really the spelling in the XSD, we should keep it.

PR #246, opened 2026-02-28
Refactor CLI into subcommands and add osa alias

seirl:
Refactored the CLI using Typer's subcommand feature. We now have `generate`, `verify`, and `kursliste` subcommands. Re-organized the main module `steuerauszug.py` to extract `run_pipeline` which gets called by the subcommands. Added an alias `osa` in `pyproject.toml` while keeping `opensteuerauszug` as the main documented binary. Adapted integration tests to match the new invocation syntax.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -127,7 +127,7 @@ You need the official Kursliste XML file for the relevant tax year.
 OpenSteuerAuszug can automatically download, prepare, and convert the latest Kursliste for you. This is the simplest method.
 
 ```bash
-opensteuerauszug kursliste fetch --year 2024
+opensteuerauszug kursliste download --year 2024
 ```
 
 This command performs the following steps:
--- a/src/opensteuerauszug/kursliste/__main__.py
+++ b/src/opensteuerauszug/kursliste/__main__.py
@@ -16,7 +16,7 @@ def main():
     pass
 
 @app.command()
-def fetch(
+def download(
     year: int = typer.Option(..., "--year", help="Tax year to download Kursliste for."),
     destination: Optional[Path] = typer.Option(None, "--destination", "-d", help="Directory to save the Kursliste XML file. Defaults to XDG data home."),
     convert: bool = typer.Option(True, "--convert/--no-convert", help="Automatically convert the downloaded XML to SQLite for faster processing.")
--- a/tests/kursliste/test_cli_download.py
+++ b/tests/kursliste/test_cli_download.py
@@ -14,7 +14,7 @@ def test_download_with_convert():
         mock_download.return_value = Path("/tmp/kursliste_2023.xml")
         mock_exists.return_value = True
 
-        result = runner.invoke(app, ["fetch", "--year", "2023"])
+        result = runner.invoke(app, ["download", "--year", "2023"])
 
         assert result.exit_code == 0
         mock_download.assert_called_once()
@@ -29,7 +29,7 @@ def test_download_no_convert():
         mock_download.return_value = Path("/tmp/kursliste_2023.xml")
         mock_exists.return_value = True
 
-        result = runner.invoke(app, ["fetch", "--year", "2023", "--no-convert"])
+        result = runner.invoke(app, ["download", "--year", "2023", "--no-convert"])
 
         assert result.exit_code == 0
         mock_download.assert_called_once()


pyproject.toml
```diff
@@ -34,8 +34,8 @@ dependencies = [
 ]
 
 [project.scripts]
-# opensteuerauszug = "opensteuerauszug.cli:app"
-opensteuerauszug = "opensteuerauszug.steuerauszug:app" # Updated entry point
+opensteuerauszug = "opensteuerauszug.cli:app"
+osa = "opensteuerauszug.cli:app" # Updated entry point
```
vroonhof, line 38: I know it happens only in the venv we are activating, but I feel this software is not used often enough to squat the three letter command space.

src/opensteuerauszug/steuerauszug.py
```diff
@@ -68,35 +67,30 @@ class LogLevel(str, Enum):
 
 default_phases = [Phase.IMPORT, Phase.VALIDATE, Phase.CALCULATE, Phase.RECONCILE_PAYMENTS, Phase.RENDER]
 
-@app.command()
-def main(
```
vroonhof, line 72: it feels weird to have a file named after the main command that isn't the actual command. we should then rename this is to pipeline.py or driver.py

docs/user_guide.md
```diff
@@ -96,7 +96,7 @@ You can specify these options multiple times to add multiple files. The files wi
 
 Example:
 ```bash
-python -m opensteuerauszug.steuerauszug input.xml --importer ibkr \
+opensteuerauszug generate input.xml --importer ibkr \
```
vroonhof, line 99: if we are going to do this we should upfate the shell scripts in scripts/... as well.

src/opensteuerauszug/cli.py
```diff
@@ -0,0 +1,103 @@
...
+    broker_name: Optional[str] = typer.Option(None, "--broker", help="Broker name from config.toml to use for this run."),
+    kursliste_dir: Optional[Path] = typer.Option(None, "--kursliste-dir", help="Directory containing Kursliste XML files."),
+    org_nr: Optional[str] = typer.Option(None, "--org-nr", help="Override the organization number used in barcodes."),
+    payment_reconciliation: bool = typer.Option(True, "--payment-reconciliation/--no-payment-reconciliation", help="Run optional payment reconciliation between Kursliste and broker evidence."),
+    pre_amble: Optional[List[Path]] = typer.Option(None, "--pre-amble", help="List of PDF documents to add before the main steuerauszug."),
+    post_amble: Optional[List[Path]] = typer.Option(None, "--post-amble", help="List of PDF documents to add after the main steuerauszug."),
+):
+    """Generates a Steuerauszug from broker data or raw XML."""
+    global_opts: GlobalOptions = ctx.obj
+    run_pipeline(
```
vroonhof, line 55: perhaps it is less of a PITA with coding agents but  I find it really annoying what we now have to add a line of boilerplate here for every flag we add. (and perhaps even two for if tit shared with the verify path).

src/opensteuerauszug/cli.py
```diff
@@ -0,0 +1,103 @@
...
+):
+    """Verifies an existing Steuerauszug XML file."""
+    global_opts: GlobalOptions = ctx.obj
+    run_pipeline(
+        input_file=input_file,
+        run_phases_input=[Phase.VERIFY],
+        raw_import=True,
+        tax_year=tax_year,
+        filter_to_period_flag=False,
+        tax_calculation_level=TaxCalculationLevel.NONE,
```
vroonhof, line 95: this is actually wrong, we want to do the tax calculations which is actually the main point of the verify phase.

seirl:
> I need to sleep over this. I like the new interface, I don't really like the new structure of the code.

Yeah that's fair, I also wasn't a fan of the output. I don't have strong opinions on this, I mostly wanted to have one binary entry point to make it easier to package. Feel free to suggest anything different!

pet-zh:
I like the idea of extracting the CLI interface from steuerauszug.py. This would allow us to build other interfaces, such as a REST interface.

vroonhof:
Quick update on this. As I wanted some changes and this likely in need of a rebase anyway I asked few AI friends for their implementation to see if there is some idea for structuring this cleanly

vroonhof:
Superseded by PR #292 . Thanks for the idea and the nudge

PR #249, opened 2026-02-28
IBKR: Import negative debit interest

pet-zh:
We only import negative debit interest (interest payable due to a negative account balance). Negative credit interest, which is charged when a currency has a negative interest rate, cannot be treated as a liability payment and could instead be classified as fees.

This depends on #235


src/opensteuerauszug/importers/ibkr/ibkr_importer.py
```diff
@@ -770,9 +770,14 @@ def should_skip_entry(entry: Any, entry_label: str) -> bool:
...
                         elif tx_type in [ibflex.CashAction.BROKERINTPAID]:
-                            # TODO: Optionally create a liabilities section.
-                            logger.warning(f"Broker interest paid for {description} is not handled for liabilities.")
-                            continue
+                            # Interst paid due to negative balance: description starting with "<CURRENCY> DEBIT INT FOR"
+                            if description.startswith(f"{currency} DEBIT INT FOR"):
+                                # Tax relevant event. Fall through to create a bank payment.
+                                pass
+                            else:
+                                # TODO: CREDIT INT is charged on positive balance and would belong to fees (not liabilities).
```
vroonhof, line 778: perhaps we could have explicit checked for CREDIT INT, but as you added a good test.

PR #254, opened 2026-03-03
chore: Extract translations from render.py

pet-zh:
This change just extracts translations from render.py (importers are not affected yet).
This change does not change any translations. Output is 1:1 equal.
Translations for other languages will be added in a follow-up change.

I have also looked into a `gettext` based approach, but this might be an overkill for our needs where translations are quite static.


pet-zh:
@vroonhof Please assign @copilot for reviewing this change.

PR #260, opened 2026-03-04
fix(schwab): handle mutation-only position histories in strict mode

### LINKED ISSUE #259: Schwab importer should handle mutation-only histories without balance snapshots
### Summary
The Schwab importer can fail strict reconciliation when a security has only mutation events in the period and no explicit `mutation=False` balance snapshot row.

### Error
`ValueError: [<position>-initial_check] Consistency check failed: No balance statement (mutation=False) found to start reconciliation.`

### Why this is valid input
A security that is opened and closed within the tax period may naturally have only transaction mutations (buys/sells/transfers) and end with zero quantity, without a separate snapshot row in imported position evidence.

### Expected behavior
Importer should tolerate mutation-only histories and synthesize boundaries from zero where needed, instead of hard-failing strict consistency at initial check.

### Actual behavior
Initial strict consistency check raises and aborts import.

### Suggested behavior
- Treat "no balance snapshot found" as a recoverable case for importer preprocessing.
- Continue with boundary synthesis using zero-assumption when no balance rows exist.

### Acceptance criteria
- Mutation-only security histories do not fail import in strict mode.
- Boundary synthesis produces stable start/end positions for these cases.
- Regression test covers mutation-only input history.

raducotescu:
Treat missing balance snapshots in initial reconciliation as a recoverable importer case and continue with synthesized boundaries.

- catch the specific no-balance initial_check error in Schwab importer
- synthesize start/end boundaries with assume_zero_if_no_balances=True
- add regression test for mutation-only security history under strict consistency

Fixes #259

### DIFF
--- a/src/opensteuerauszug/core/position_reconciler.py
+++ b/src/opensteuerauszug/core/position_reconciler.py
@@ -86,91 +86,36 @@ class PositionReconciler:
             )
             current_quantity = Decimal("0")
             current_date = self.sorted_stocks[0].referenceDate
-            is_consistent = True
-            log_messages = []
+            start_idx = 0
+        else:
+            start_balance_event = self.sorted_stocks[first_balance_idx]
+            current_quantity = start_balance_event.quantity
+            current_date = start_balance_event.referenceDate
+            start_idx = first_balance_idx + 1
+            logger.debug(f"{log_prefix} Starting consistency check from initial balance on {current_date}. Initial Qty: {current_quantity} ({start_balance_event.balanceCurrency}).")
 
-            for stock_event in self.sorted_stocks:
-                event_date = stock_event.referenceDate
-                if event_date < current_date:
-                    msg = f"ERROR: {log_prefix} Stock events appear out of order: event on {event_date} after processing {current_date}. Aborting."
-                    logger.error(msg)
-                    log_messages.append(msg)
-                    is_consistent = False
-                    break
-
-                if stock_event.mutation:
-                    delta_quantity = stock_event.quantity
-                    current_quantity += delta_quantity
-                    logger.debug(
-                        f"{log_prefix} Date: {event_date}. Applying mutation: "
-                        f"Name='{stock_event.name or 'N/A'}', Qty Change={delta_quantity}. "
-                        f"New Qty: {current_quantity}."
-                    )
-                else:
-                    # Defensive branch for future sorting/inputs where balances appear unexpectedly.
-                    if current_quantity != stock_event.quantity:
-                        msg = (
-                            f"ERROR: {log_prefix} Mismatch on {event_date}! "
-                            f"Calculated Qty: {current_quantity}, Reported Qty in statement: {stock_event.quantity}."
-                        )
-                        logger.error(msg)
-                        log_messages.append(msg)
-                        is_consistent = False
-                    current_quantity = stock_event.quantity
-
-                current_date = event_date
-
-            if current_quantity != Decimal("0"):
-                msg = (
-                    f"ERROR: {log_prefix} Mutation-only consistency check failed: "
-                    f"final quantity is {current_quantity}, expected 0."
-                )
-                logger.error(msg)
-                log_messages.append(msg)
-                is_consistent = False
-
-            if is_consistent:
-                logger.info(f"{log_prefix} Mutation-only consistency check finished successfully.")
-            elif raise_on_error:
-                error_summary = f"[{self.identifier}] Position reconciliation failed. Log:\n" + "\n".join(log_messages)
-                raise ValueError(error_summary)
-
-            return is_consistent, log_messages
-
-        start_balance_event = self.sorted_stocks[first_balance_idx]
-        current_quantity = start_balance_event.quantity
-        current_date = start_balance_event.referenceDate
-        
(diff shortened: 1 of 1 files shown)


src/opensteuerauszug/importers/schwab/schwab_importer.py
```diff
@@ -118,18 +118,31 @@ def _reconcile_and_ensure_boundary_stocks_for_position(
...
-            raise_on_error=self.strict_consistency
-        )
+        no_balance_guard = "No balance statement (mutation=False) found to start reconciliation."
+        try:
+            is_consistent_initial, _ = initial_reconciler.check_consistency(
+                print_log=True,
+                raise_on_error=self.strict_consistency
+            )
+        except ValueError as exc:
+            if no_balance_guard in str(exc):
```
vroonhof: let's not do this having a caught exception in the normal flow is already bad, but using string compare on the type is worse,

src/opensteuerauszug/importers/schwab/schwab_importer.py
```diff
@@ -118,18 +118,31 @@ def _reconcile_and_ensure_boundary_stocks_for_position(
...
         initial_reconciler = PositionReconciler(list(initial_pos_stocks), identifier=f"{current_identifier}-initial_check")
-        is_consistent_initial, _ = initial_reconciler.check_consistency(
-            print_log=True,
-            raise_on_error=self.strict_consistency
-        )
+        no_balance_guard = "No balance statement (mutation=False) found to start reconciliation."
+        try:
+            is_consistent_initial, _ = initial_reconciler.check_consistency(
+                print_log=True,
+                raise_on_error=self.strict_consistency
```
vroonhof: rather than catching the exception, how about adding support for assume_zero_if_no_balances to this method as well (and then as a consistency check, check that starting with gives zero at the end)

raducotescu: Switched to a better implementation in [13e9d5](https://github.com/vroonhof/opensteuerauszug/pull/260/commits/713e9d56cf040d361a4932bd4c77b9c1e4cf1340).

src/opensteuerauszug/core/position_reconciler.py
```diff
@@ -65,11 +73,69 @@ def check_consistency(self, print_log: bool = False, raise_on_error: bool = Fals
...
+            logger.warning(
+                f"{log_prefix} No balance statement found. "
+                "Assuming initial quantity of 0 for mutation-only consistency check."
+            )
+            current_quantity = Decimal("0")
+            current_date = self.sorted_stocks[0].referenceDate
+            is_consistent = True
+            log_messages = []
+
+            for stock_event in self.sorted_stocks:
```
vroonhof: I think this block is now indenteed too far...   if I am reading it correctly this now all in the if first_balance_idx == -1 block

raducotescu: I think it's correct; before we were exiting the block on inconsistencies, now we try to reconcile. The provided test should clarify this.

vroonhof: ah... the whole logic got duplicated.. that should not be necessary. I think we can just flag we did this and run the formal block with a check at the end. I can fix htis in the merge it if you want (just not now).

raducotescu: Up to you, my Python skills are rusty – haven't written Python in 14-15 years now.

I have however found the issues when I was running the Schwab importer and I wanted to contribute fixes for the persons who might encounter the same problems.

Locally I have patched all the bugs I encountered and got my PDFs (thank you!).

If you want nicer fixes that what I provided, then by all means implement them.

vroonhof:
On Fri, Mar 6, 2026 at 11:41 AM Radu Cotescu ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/opensteuerauszug/core/position_reconciler.py
> <https://github.com/vroonhof/opensteuerauszug/pull/260#discussion_r2895074693>
> :
>
> > +                msg = f"{log_prefix} Consistency check failed: No balance statement (mutation=False) found to start reconciliation."
> +                logger.error(msg)
> +                if raise_on_error:
> +                    raise ValueError(msg)
> +                return False, [msg]
> +
> +            logger.warning(
> +                f"{log_prefix} No balance statement found. "
> +                "Assuming initial quantity of 0 for mutation-only consistency check."
> +            )
> +            current_quantity = Decimal("0")
> +            current_date = self.sorted_stocks[0].referenceDate
> +            is_consistent = True
> +            log_messages = []
> +
> +            for stock_event in self.sorted_stocks:
>
> Up to you, my Python skills are rusty – haven't written Python in 14-15
> years now.
>
> I have however found the issues when I was running the Schwab importer and
> I wanted to contribute fixes for the persons who might encounter the same
> problems.
>

Thanks for doing that, it is very valuable.. The software intentionally
fails when it encounters a situation that we have not tested before so that
we have real cases to test it with.


> Locally I have patched all the bugs I encountered and got my PDFs (thank
> you!).
>
> If you want nicer fixes that what I provided, then by all means implement
> them.
>
As I said that is OK. (Sadly due to github permissions I cannot pay with my
LLM credits trivially to do it for you, I need to take over you).

Jan

PR #266, opened 2026-03-05
Add translations for EN, FR, IT

pet-zh:
Some minor layout adjusts were necessary for proper line breaks.


pet-zh:
@vroonhof Please assign @copilot for review here.

PR #267, opened 2026-03-05
Add '--language' CLI parameter

pet-zh:
This allows config-less execution.


src/opensteuerauszug/steuerauszug.py
```diff
@@ -94,6 +94,7 @@ def main(
     override_configs: List[str] = typer.Option(None, "--set", help="Override configuration settings using path.to.key=value format. Can be used multiple times."),
     kursliste_dir: Optional[Path] = typer.Option(None, "--kursliste-dir", help="Directory containing Kursliste XML files for exchange rate information. Defaults to 'data/kursliste' in CWD or XDG data home."),
     org_nr: Optional[str] = typer.Option(None, "--org-nr", help="Override the organization number used in barcodes (5-digit number)"),
+    language: Optional[str] = typer.Option(None, "--language", help="Override output document language ('de', 'fr', 'it', 'en')."),
```
vroonhof, line 97: I don't think is necessary... this shouldalready be settable with "--set language=xx" (all config setting are settable from the command line).

vroonhof (changes requested):

src/opensteuerauszug/steuerauszug.py
pet-zh: Thanks for pointing out - you are right. I have missed that one.

While testing it, I have noticed that it does not work as expected. Fixed in #271.

This allows then overriding the language: `--set general.language=fr`

PR #269, opened 2026-03-06
Refine type hinting and attribute handling for payment subtypes

vroonhof:
This PR refines the dynamic type hinting and attribute extraction logic for payment objects in `KurslisteTaxValueCalculator`. Previously, `# TODO: Should the typing be smarter?` was noted due to the varying existence of fields like `gratis`, `capitalGain`, and `exDate` across payment subtypes. 

The approach introduces `Protocol` classes to structurally outline expected data, typing the primary iteration loop correctly, and safely type-guarding access to optional fields without causing `AttributeError` regressions at runtime or breaking older Python versions (where data-only `@runtime_checkable` `isinstance()` fails).

---
*PR created automatically by Jules for task [14699823914039545263](https://jules.google.com/task/14699823914039545263) started by @vroonhof*

### DIFF
--- a/src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
+++ b/src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
@@ -1,6 +1,7 @@
 from decimal import Decimal
-from typing import Optional, List, Set
+from typing import Optional, List, Set, Protocol, Any, cast
 import logging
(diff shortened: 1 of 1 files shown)


vroonhof:
types here don't seem to map to something semantic in the kursliste. not sure it ads value.

PR #275, opened 2026-03-06
fix(calculator): suppress Kursliste warnings for zero-balance options

aureld:
## Summary

- Options (derivatives) are never listed in the Swiss ESTV Kursliste
- When an option position is **fully closed before year-end** (zero closing balance), its tax value is definitively 0 — generating a critical warning is misleading and blocks users from completing their tax return
- Extends the existing rights-issue suppression logic to also cover `OPTION` securities with zero closing balance: suppresses the warning and sets an explicit **0 CHF** value (with correct exchange rate) instead of marking the entry as `undefined` / "n.v."

## Motivation

Users with short options that were opened and closed within the tax year (e.g. short puts bought back before expiry) would receive one critical warning per option contract — easily 70+ warnings for an active options strategy — even though those positions have zero tax relevance at year-end.

### DIFF
--- a/src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
+++ b/src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
@@ -82,6 +82,7 @@ class KurslisteTaxValueCalculator(MinimalTaxValueCalculator):
             self.kursliste_manager = exchange_rate_provider.kursliste_manager
         self.flag_override_provider = flag_override_provider
         self._current_kursliste_security = None
+        self._current_security_is_zero_balance_option = False
         self._missing_kursliste_entries = []
         self._stock_split_warnings: List[dict] = []
         self._previous_year_exdate_warnings = []
@@ -136,6 +137,7 @@ class KurslisteTaxValueCalculator(MinimalTaxValueCalculator):
 
     def _handle_Security(self, security: Security, path_prefix: str) -> None:
         self._current_kursliste_security = None
+        self._current_security_is_zero_balance_option = False
 
         if not self.kursliste_manager:
             super()._handle_Security(security, path_prefix)
@@ -180,8 +182,9 @@ class KurslisteTaxValueCalculator(MinimalTaxValueCalculator):
                 else security.securityName
             )
 
-            # Check if this is a rights issue that we should ignore if not found
+            # Check if this is a rights issue or zero-balance option that we should ignore if not found
             is_rights = security.is_rights_issue
+            is_option = security.securityCategory == "OPTION"
             closing_balance = Decimal("0")
 
             if security.taxValue:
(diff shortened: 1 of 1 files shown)


vroonhof:
We are skirting into fill-in tax calculation mode here.

Is there some authoriative source that says that options do not cause tax events during the year. In particular short ones where you write and so get the premium (BTW interesting most of the kursliste consists of derivatives, just none of them are options).

pet-zh:
@aureld Thanks for your PR. Can you please give #224 a try (you need to cherry pick or rebase / merge main)? That one should improve this as well in a more generic way. We still need to decide which way we want to go, so please keep this one open.

vroonhof:
I think we would need a mix. because even if the value balance the end of the period is zero, there could have been a tax implication due to balances being non-zero during the year (e.g. if it was an accumulating fund and it had a tax event during that time).

I think this PR is doing two things
- stop warning about kursliste from zero balances at the end
- assert that for options the EOY value is the the only think that mattters (just like for rights issues). 

Ideally we would have an authorative source (even this always how the tax office does it I do it manually would hapen)

Am I actually surprised the kursliste has tons of structured products but no options, because at least for traded options there is a reasonable objective marketprice your could use.

vroonhof:
Just putting it here for posterity. This the best information that seems to exist online

From [Besteuerung von Obligationen, Derivatenund kombinierten Produkten (ESTV, Bern 2023)](https://www.estv2.admin.ch/stp/ds/f-finanzinstrumente-de.pdf)

```
3.4.3 Direkte Bundessteuer sowie kantonale und
kommunale Einkommenssteuern
Gemäss Rechtsprechung des Bundesgerichts sind Gewinne aus Termingeschäften steuerlich gleich
zu behandeln wie solche aus Kassageschäften und stellen deshalb Kapitalgewinne dar.5
 Soweit Kapitalgewinne durch das Gesetz nicht ausdrücklich erfasst werden, bleiben solche aus Termingeschäften (Futures) und Optionsgeschäften sowohl beim Bund (Art. 16 Abs. 3 DBG) als auch in den Kantonen und Gemeinden (Art. 7 Abs. 4 Bst. b StHG) für die Belange des Privatvermögens steuerfrei.
Entsprechende Verluste im Privatvermögen sind dafür auch nicht abzugsfähig. 
```

note that the Optionspremium is not directly mentioned. Just "gains".  

This article by [Marcel Jung, Tax advisor](https://froriep.com/wp-content/uploads/Marcel_Jung_Taxation_of_structured_products.pdf) does mention it explicit. But of course it is not an official communication.

vroonhof:
Given this reuses existing logic it does not add any new debt that would make https://github.com/vroonhof/opensteuerauszug/pull/224 harder to reason about. so for now I am applying it.

PR #277, opened 2026-03-09
Add VT to example security indentifiers.

jdoerrie:
See https://www.ictax.admin.ch/extern/en.html#/security/US9220427424/20251231.


pet-zh:
@jdoerrie Just wondering: what is your motivation for adding this? What does it improve?

jdoerrie:
I hold VT in my Schwab account, and thus need to declare my holdings and received dividends. Since Schwab only deals with ticker symbols, the mapping to ISIN or Valor number needs to be made explicitly: https://github.com/vroonhof/opensteuerauszug/blob/main/docs/importer_schwab.md#mapping-symbols-to-kurslistevalor

While I certainly can add the relevant line in my local checkout, I figured holding VT (and chill) is common enough that it makes sense to include it by default.

Issue #282, opened 2026-03-11
What to do when the Stock is not on Kurslist?


smarley2:
Hi, I have the following US stocks that are not present on the kurslist.
How can I manually fix it?
Sorry I did not found this troubleshooting on the instructions.

<img width="899" height="199" alt="Image" src="https://github.com/user-attachments/assets/adc74bec-84b5-4b8e-a502-1a32b4d966c2" />

<img width="299" height="652" alt="Image" src="https://github.com/user-attachments/assets/48cf2d7e-c49d-47a4-9afc-4f5dc09469e9" />

<img width="333" height="116" alt="Image" src="https://github.com/user-attachments/assets/382adb6d-7b6a-410e-88f3-5f7125356e15" />

vroonhof:
The correct thing to do is not to manually fix, but instead to request to add it to the kursliste. For one off requests you can just send a friendly email to dvs@estv.admin.ch.  if you have a lot you can use the web form (https://www.ictax.admin.ch/extern/docs/en/ictax_help.pdf) but that also requires your email to be allowlisted first.

Adding them to the kursliste takes a few days and then you have to wait till the next update  bundle to be availble

vroonhof:
BTW if actually upload a steuerausweis with missing fiels to the Kursliste into the Zurich tax software it allows you to manualy fill in the fields but you will not get no assistance to calculate the payments, kursliste option is much better

smarley2:
> BTW if actually upload a steuerausweis with missing fiels to the Kursliste into the Zurich tax software it allows you to manualy fill in the fields but you will not get no assistance to calculate the payments, kursliste option is much better

ok, thank you for the advice, I will request them to add this stocks to the list as you mention.

smarley2:
fyi they added to the website pretty fast, in less than 1h.
Send the email to wefin.dvs@estv.admin.ch

pet-zh:
@vroonhof I would suggest to reopen this and add clear instructions in the ouput PDF how to handle by the user (as you described above). That will also help others.

smarley2:
Ok, make sense, I also got that they update weekly on Friday evening until calendar week 14. After that is every two weeks until calendar week 28 and 4 weeks after.

smarley2:
Described here: https://github.com/vroonhof/opensteuerauszug/pull/284

PR #284, opened 2026-03-11
Document stock request process for Kursliste

smarley2:
This PR adds documentation explaining how users can request the inclusion of missing securities in the ESTV Kursliste.
Described on the issue https://github.com/vroonhof/opensteuerauszug/issues/282

Sometimes US stocks or less common securities are not yet available in the official list. The documentation now explains that users can email ESTV with the ISIN to request inclusion, which is typically processed quickly and then appears in the next XML update.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -170,9 +170,7 @@ If a stock cannot be found in the Kursliste you can request the addition of the 
 
 Send an email with the ISIN and company name to:
 
-- wefin.dvs@estv.admin.ch  
-- evestv.admin@zh.ch  
-- dvs@estv.admin.ch  
+- dvs at estv.admin.ch  
 
 Example email:
 


docs/user_guide.md
```diff
@@ -164,6 +164,37 @@ Place the downloaded Kursliste XML file(s) and generated SQLite database(s) into
...
 
+### Stock not found in the Kursliste
+
+If a stock cannot be found in the Kursliste you can request the addition of the security directly from the Swiss Federal Tax Administration.
+
+Send an email with the ISIN and company name to:
+
+- wefin.dvs@estv.admin.ch  
+- evestv.admin@zh.ch  
+- dvs@estv.admin.ch
```
vroonhof: I think we should just list one afress. And I think DVS is the right one.

PR #288, opened 2026-03-14
fix: treat positive withholding tax entries as reversals, not income

agners:
When IBKR corrects withholding from 30% to 15%, it books a positive "Withholding Tax" entry to reverse the initial deduction. The importer was setting grossRevenueB on such entries, which caused _accumulate_broker() to fall through and add the reversal amount to the broker dividend total instead of netting it against withholding.

Fix: set nonRecoverableTaxAmountOriginal = -amount for positive withholding entries so the reconciliation calculator reduces the withholding total and skips the dividend accumulation path.

This fixes the Payment Reconciliation in my report.


vroonhof:
Thank you.

PR #289, opened 2026-03-14
Add support for unsettled trades reconciliation for Schwab converter.

larswenker:
# Fix: Support for Unsettled Cash Reconciliation in Schwab Importer

## Summary
This PR introduces a robust mechanism for handling cash reconciliation discrepancies in Schwab accounts caused by settlement lags (sweep lags). It also enhances the core reconciliation infrastructure to provide better diagnostic information when mismatches occur.

## The Problem
Schwab statement balances often reflect "settled cash" rather than "real-time" balances. For example, a stock sale on September 29th might not "sweep" into the cash balance until October 2nd.

In the previous version (main), the importer would calculate a balance of $X (including the sale) but see a statement balance of $0 (pre-sweep). This resulted in:
1.  **Immediate Failure**: Reconciliation would fail on the boundary date (e.g., Oct 1st) because the trade was "invisible" to the statement.
2.  **Cascading Discrepancies**: The reconciler's default behavior is to reset its internal counter to the reported statement balance whenever a mismatch is found. This means that a single unresolved discrepancy creates a permanent offset for the rest of the year, causing even perfectly correct subsequent months to fail reconciliation.

### 1. Iterative Reconciliation Loop (Schwab Importer)
The SchwabImporter now includes specialized logic to resolve these settlement timing differences:
- **Trade-Back Matching**: When a cash discrepancy is detected, the importer looks back at trades within a 5-day window. If the sum of these trades exactly accounts for the discrepancy, the statement balance is adjusted to include the "in-flight" cash.
- **Iterative Re-reconciliation**: To prevent an early discrepancy from affecting later months, the importer now fixes one mismatch at a time and then **re-runs the entire reconciliation process** from the beginning. This ensures that once a mismatch is explained, the cumulative balance is corrected for all subsequent dates in the period.
- **File**: src/opensteuerauszug/importers/schwab/schwab_importer.py

### 2. Structured Reconciliation Diagnostics (Core)
The core reconciliation logic was refactored to provide more actionable error reporting:
- **Structured Results**: `PositionReconciler.check_consistency` now returns a list of ReconciliationMismatch objects instead of simple log messages.
- **Detailed Error Messages**: When a mismatch cannot be resolved, the resulting `ValueError` now includes exact "Calculated vs Reported" dollar amounts, allowing users to immediately see the size and direction of the discrepancy.
- **File**: src/opensteuerauszug/core/position_reconciler.py

### 3. Enhanced Test Coverage
New stress tests have been added to ensure stability:
- **Consecutive Lag Verification**: Ensures that multiple settlement lags occurring at different points in the year (e.g., Q3 and Q4) are all resolved correctly.
- **Boundary Testing**: Validates that trades occurring exactly on the settlement window limit are handled correctly.
- **Files**: tests/importers/schwab/test_schwab_importer.py and tests/core/test_position_reconciler.py

## Verification Results
- **Automated Tests**: All 55 unit tests passed.
- **Real-world Data**: Successfully generated tax statements for my real-world Schwab equity account where I had several trades right before a report (09/29 and on 12/31) which was broken before.

### DIFF
--- a/src/opensteuerauszug/importers/schwab/schwab_importer.py
+++ b/src/opensteuerauszug/importers/schwab/schwab_importer.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import pprint
 from typing import List, Dict, Any, Optional, Tuple
 import os
@@ -14,10 +15,12 @@ from .position_extractor import PositionExtractor
 from .transaction_extractor import TransactionExtractor
 from opensteuerauszug.util.date_coverage import DateRangeCoverage
 from collections import defaultdict
-from opensteuerauszug.core.position_reconciler import PositionReconciler, ReconciledQuantity
+from opensteuerauszug.core.position_reconciler import PositionReconciler, ReconciledQuantity, ReconciliationMismatch
 from opensteuerauszug.config.models import SchwabAccountSettings # Add this
 
 logger = logging.getLogger(__name__)
+
+UNSETTLED_WINDOW_DAYS = 5
 
 # Placeholder import for TransactionExtractor (to be implemented)
 # from .TransactionExtractor import TransactionExtractor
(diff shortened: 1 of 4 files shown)


vroonhof:
Ah this is interesting. In particular seems it seems (rather counterintuitively perhaps) that for dividends the tax treatment is exactly the other way around . Thanks for contributing. I will look at this after actually doing my taxes this weekend ;-)

src/opensteuerauszug/importers/schwab/schwab_importer.py
```diff
@@ -106,6 +109,71 @@ def _determine_synthesized_stock_currency(
...
+                continue
+
+            recent_sum = sum(s.quantity for s in recent_mutations)
+
+            if recent_sum == -discrepancy:
+                logger.info(
+                    f"[{identifier}] Unsettled cash detected: reported {mismatch.reported_quantity} "
+                    f"on {balance_date}, calculated {mismatch.calculated_quantity}. "
+                    f"Explained by {len(recent_mutations)} recent trade(s) summing to {recent_sum}. "
+                    f"Adjusting reported balance to {mismatch.calculated_quantity}."
```
vroonhof, line 156: So this means that if you have a sale resulting on X on 31.12 that settles on 01.01 the balance we report int he steuerauzug is settled+X. I think this is (counterintuitively= wrong (for example in the case of ex- and pay dates for dividends across a boundary this does not happend=  and it makes it harder to manually check the balances. 

I agree that feels it should work like this, but it apparently doesn't. That makes your fix harder to implement,,,  (I think you should then report X has a "claim" on your broker but I don't think anybody does it,). would be nice to find an official guidance.

vroonhof: Perhaps adjusting the date of the cash transaction would be better...

larswenker: "So this means that if you have a sale resulting on X on 31.12 that settles on 01.01 the balance we report int he steuerauzug is settled+X"

I think this is how it should work, no? In my case, I sold X stock on 12/31 (let's say for $10K), but my report balance says 0 on Jan 1. I don't think reporting a 0 balance would be correct, it should still count towards my overall wealth as cash. 

The sweep of balance during the settlement period feels more like an implementation detail. If we had looked at this set of transactions later we'd just have listed the trade date as 12/31, this artifact/edge case only happens because we happen to check a balance during the settlement date.

I'm not sure I understand the comparison to dividends, can you explain what you'd expect the behavior to be instead in this case?

vroonhof: The comparison with dividends is because if a stock goes ex just before the end of year I would think is effectively creates a claim you have on the issuer which is something of value so your wealth is increases by the value of that claim. This not what happens. The wealth increase and thus the corresponding tax event is booked on the pay date.

vroonhof: I tried looking up how this supposed to work. There are some hints that one should create a separate bank account entry for unsettled funds but I have not seen any real examples of that actually happening.

vroonhof: we do need to find a solution, because I presume we just throw a hard error at the moment as it is (correctly) flagging an issue in the interpretation of the figures.

vroonhof:
Hi @larswenker would you be willing to check https://github.com/vroonhof/opensteuerauszug/pull/345 with your data? it assumes settlement is fixed so you don't have to search for it..

vroonhof:
I have merged #345 though I don't have a good way to test it until I next do a trade on my schwab account

PR #293, opened 2026-03-15
Added Swisquote Importer

bensiegit:
**Add Swissquote German CSV importer**
Adds a SwissquoteImporter class supporting the German-language CSV export from Swissquote Bank, tested against a real 2025 export (XSD validation passes, PDF renders successfully).
Changes:

- importers/swissquote/swissquote_importer.py — full class-based importer with cp1252 encoding, timestamp-aware date parsing, 28 mapped German transaction types, and opening balance computation from full history CSV

- config/models.py — added SwissquoteAccountSettings, updated union and literal types

- steuerauszug.py — added SWISSQUOTE to ImporterType enum and dispatch block

Known limitations: single-currency CHF account only; securities lending without ISIN skipped; delisted securities (CS Group, BTC/ETH) show negative balances with no tax impact.


docs/importer_swissquote.py
```diff
@@ -0,0 +1,382 @@
+"""
+Swissquote CSV importer for opensteuerauszug.
```
vroonhof: Why is there a whole copy of the importer in docs. I have heard of "code is the best documentation" but this is pushing it :)

docs/importer_swissquote.py
```diff
@@ -0,0 +1,382 @@
...
+    value = value.strip()
+    for fmt in _DATE_FORMATS:
+        try:
+            return _dt.datetime.strptime(value, fmt).date()
+        except ValueError:
+            continue
+    raise ValueError(f"Cannot parse date: {value!r}")
+
+
+def _detect_delimiter(filepath: Path) -> str:
```
vroonhof: Are swissquoute really that inconsistent in their choice of delimiter?

docs/importer_swissquote.py
```diff
@@ -0,0 +1,382 @@
...
+
+# ---------------------------------------------------------------------------
+# Core parser
+# ---------------------------------------------------------------------------
+
+def parse_swissquote_csv(filepath: Path) -> Iterator[SwissquoteTransaction]:
+    """
+    Parse a Swissquote German CSV export and yield SwissquoteTransaction
+    objects one per row.  Rows with unrecognised transaction types are
+    logged as WARNINGs and skipped — add the label to TRANSACTION_TYPE_MAP.
```
vroonhof: I think a better option for software like this is to throw an exception when encountering unknown data.

docs/importer_swissquote.py
```diff
@@ -0,0 +1,382 @@
...
+
+# ---------------------------------------------------------------------------
+# Dividend / withholding-tax pairing
+# ---------------------------------------------------------------------------
+
+def pair_dividends_and_taxes(
+    transactions: list[SwissquoteTransaction],
+) -> list[tuple[SwissquoteTransaction, Optional[SwissquoteTransaction]]]:
+    """
+    Swissquote emits dividend income and the corresponding withholding tax
```
vroonhof: All our brokers seem to that. This logic is now repeated. Reminder to self to think about centralizing

vroonhof (commented):
Does this do the right thing with Steuerrückbehalt USA? I think this the first swiss broker? 

Some simple documentation seems useful. Is this expecting the export to contain all transactions since the dawn of time/the account opening? Is there anything we can do to create some safety here?

How much manual inspection/review of the output or the code was there?

docs/importer_swissquote.py
bensiegit: Yes that makes no sense.

pet-zh:
@bensiegit Do you plan to get this PR ready to merge? I assume many Swissquote clients would appreciate this. There is also a paid tool popping up: #413.

PR #318, opened 2026-03-26
Add Docker support for simplified usage and setup

felldo:
First of all, thank you for creating this.

I'm not gonna lie, but the setup was kind of a PITA and I'm a Software Engineer. There is just so much documentation which often also seems useless for most people (end users) and its gets very confusing. Also some commands in the documentation aren't working/are missing arguments. Like, I had to add the date cli arguments when using the tool for IBKR, which were missing in the docs. So the overall experience wasn't as smooth as it could be (The first time I tried using this I actually gave up, thought I would just use datalevel for this year, but I gave it another shot).

Therefore, I added a very simple docker setup which allows everyone to run this tool without struggling and installing a whole development environment. For now I can only confirm that this works for IBKR. The docker files may have to be adjusted for proper support for other brokers.

I would also advise you to aim to enhance the docker setup instead of how it currently works and make this the primary way of running this tool. For end users it makes everything just so much easier. From configuration to execution as well as removal of everything at the end and updating for the next years tax period. No need to store anything on the OS level like the kursliste. Just delete the docker volume and that's it.

For now this is a basic docker setup but I might contribute more to it when your approve the overall switch to it.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+FROM python:3.14
+
+RUN apt-get update && apt-get install -y git && \
+    pip install --upgrade pip
+
+RUN git clone https://github.com/vroonhof/opensteuerauszug.git
+
+WORKDIR /opensteuerauszug
+
+RUN python3 -m venv .venv && \
+    .venv/bin/pip install .[dev] && \
+    .venv/bin/pip install git+https://github.com/vroonhof/pdf417-py.git
+
+RUN mkdir -p /data
+
+ARG YEAR=2024
+ENV YEAR=${YEAR}
+ENV IMPORTER=ibkr
+
+RUN .venv/bin/opensteuerauszug kursliste download --year ${YEAR}
+
+ENV EXTRA_ARGS=""
+
+CMD ["/bin/sh", "-c", ".venv/bin/opensteuerauszug process --importer ${IMPORTER} /data/input.xml --config /data/config.toml --output /data/tax_statement_${YEAR}.pdf --period-from ${YEAR}-01-01 --period-to ${YEAR}-12-31  \"${EXTRA_ARGS}\""]
--- a/README.md
+++ b/README.md
@@ -18,6 +18,24 @@ For more information on required due diligence see the the user guide.
 
 After installing, see the [User Guide](docs/user_guide.md) for preparation steps, how to get broker data and run the command.
 
+#### Easy usage with Docker
+
+Instead of installing python and the dependencies, you can install [Docker](https://docs.docker.com/engine/install/). It makes the usage much easier and requires minimal knowledge of programming languages and their ecosystem.
+
+After installing Docker, go into the ``./docker`` directory in this project within your terminal, optionally create a config.toml file (check [User Guide](docs/user_guide.md) if you need one) and put it into the `./docker/data/` directory. Create the directory if it doesn't exist.
+
+Now generate the data from your broker. Check the usage guide for details. Rename the xml file to ``input.xml`` and put it into the `./docker/data/` directory. Now you should have the following files in the `./docker/data/` directory.
+
+Open the ``./docker/docker-compose.yml`` file (or ``docker-compose.yml`` once you are in the ``./docker`` directory) and configure it to your needs. Adjust the year, change the importer if you need to, and optionally pass some additional arguments to the tool.
+
+Now run the following command in your terminal within the ``./docker`` directory to start the docker container and generate the PDF:
+
+```bash
+docker compose up --build
+```
+
+The PDF will be generated in the `./docker/data/` directory and will be named ``tax_statement_<YEAR>.pdf``.
+
 ### Verifying an Existing Steuerauszug
 
 The tool can also be used to cross check and existing existing Steuerauszug (eCH-0196 XML). See See [verify instructions](docs/verify_existing.md).
@@ -151,7 +169,7 @@ This project includes utility scripts for development and data management. For d
 
 ### Code quality and AI usage.
 
-This project exists in part for me to try out AI based coding outside of $WORKPLACE to try out various tools. As a result code quality and style is inconsistent and contains various AIisms. Code has been cleaned-up, reviewed and controlled by me where it matters.
+This project exists in part for me to try out AI based coding outside of $WORKPLACE to try out various tools. As a result code quality and style is inconsistent and contains various AIisms. Code has been cleaned-up, reviewed, and controlled by me where it matters.
 
 That said all mistakes, hallucinations etc are probably mine.
 
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,11 @@
+services:
+  opensteuerauszug:
+    build:
+      context: .
+      args:
+        - YEAR=2025
+    volumes:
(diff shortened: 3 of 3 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN apt-get update && apt-get install -y git && \
+    pip install --upgrade pip
+
+RUN git clone https://github.com/vroonhof/opensteuerauszug.git
+
+WORKDIR /opensteuerauszug
+
+RUN python3 -m venv .venv && \
+    .venv/bin/pip install .[dev] && \
```
pet-zh, line 11: I think you do not even need to install [dev]. Just do: `.venv/bin/pip install .`

docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    .venv/bin/pip install git+https://github.com/vroonhof/pdf417-py.git
+
+RUN mkdir -p /data
+
+ARG YEAR=2024
```
pet-zh, line 16: Probably we can default this to the previous year always?

docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ENV YEAR=${YEAR}
+ENV IMPORTER=ibkr
+
+RUN .venv/bin/opensteuerauszug kursliste download --year ${YEAR}
+
+ENV EXTRA_ARGS=""
+
+CMD ["/bin/sh", "-c", ".venv/bin/opensteuerauszug process --importer ${IMPORTER} /data/input.xml --config /data/config.toml --output /data/tax_statement_${YEAR}.pdf --period-from ${YEAR}-01-01 --period-to ${YEAR}-12-31  \"${EXTRA_ARGS}\""]
```
pet-zh, line 24: You can pass `--tax-year` instead of `--period-from` and `--period-to`.

docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+WORKDIR /opensteuerauszug
+
+RUN python3 -m venv .venv && \
+    .venv/bin/pip install .[dev] && \
+    .venv/bin/pip install git+https://github.com/vroonhof/pdf417-py.git
+
+RUN mkdir -p /data
+
+ARG YEAR=2024
+ENV YEAR=${YEAR}
```
pet-zh, line 17: In general, it would be nice to have a more dynamic setup of arguments so that anything could be passed to opensteuerauszug. That would allow re-using the parameter docs. Then docker would basically only streamline the setup, but the usage would stay the same.

pet-zh (commented):
Thanks for your idea. In general, I like the idea of streamlining the setup. This is important for allowing a broader adoption.

vroonhof:
yes it true the documentation is a mess. it was orginally target more at developers can do with an audit. It also been recently through the work of others that using it became a simple two command setup. Setting up the kursliste was messy etc.
also iterative development have meant stupid thinks like the tax year argument having two different arguments for the two subcommands

To be frank I think as is your docker setup is at risk of growing as complex as the full command set once it becomes than just for you, so it not 100% clear it is actually easier.

If we allow third party wrappers I think using 'uv' may ben be simpler. if you have uv I think the command can just be

uvx --from 'https://github.com/vroonhof/opensteuerauszug.git' opensteuerauszug ...

I have an old protype that adds an html interface and runs everything under wasm, but there is a point where you will just notice the value add that datalevel is giving you. 

at this stage I also believe we benefited a lot from having a bit more "developer" setup because open source has truly worked as many people with many different edge cases in their data have contributed fixes.

docker/Dockerfile
felldo: This is just a placeholder. The actual value will be passed from the docker compose file

docker/Dockerfile
felldo: This was my plan. But before doing anyone is doing a complete rework, I wanted merge the "base" idea / get the approval

felldo:
> To be frank I think as is your docker setup is at risk of growing as complex as the full command set once it becomes than just for you, so it not 100% clear it is actually easier.

Ideally, we should be able to replace the CLI arguments with environment variables. The main goal here is to provide a user-friendly tool for everyone in Switzerland. Right now, this is not the case. I've read many times on Reddit that people have used this tool and, if they were successful, they pretty much all said that it wasn't easy to use.
The target group is end users. Most of them will probably not contribute code, but at most they will create an issue for something. The documentation contains examples of how to download the Kursliste, convert it, and so on. In my opinion, this is pretty much irrelevant for 99% of end users. So they will never need to execute these commands. They just want a working tool. However, it's good that this is documented for developers who may need to debug something.

If we wanted to we could also move the `config.toml` file right into the docker compose to have a central configuration file.

> If we allow third party wrappers I think using 'uv' may ben be simpler. if you have uv I think the command can just be

I have no idea what this is, but from what I've read, Python would still be required to be installed. I think that, with this solution, you would always need to re download the Kursliste and convert it. This isn't very useful when you try to configure the tool for the first or subsequent uses. With the Docker setup, however, this is not the case due to the way the image works. They have steps/checkpoints, so not everything has to be rebuilt, and it should be easy to adapt the configuration to your needs.

> I have an old protype that adds an html interface and runs everything under wasm, but there is a point where you will just notice the value add that datalevel is giving you.

Of course, a website is always the preferred option for end users because it's simple and only requires a browser. However, as it contains sensitive user data, I don't think users would want to use it. What exactly does Datalevel provide that this does not? Except for ease of use. Ultimately, both solutions should generate a correct statement that can be uploaded to the tax software.

> at this stage I also believe we benefited a lot from having a bit more "developer" setup because open source has truly worked as many people with many different edge cases in their data have contributed fixes.

Maybe, but at what costs? That every normal user does not use this because it looks like rocket science to setup?
In the end everything could stay like it is right now. There are just a few adjustments we need to make. For example, CLI arguments should be able to be substituted by environment variables, and the latter should take precedence over the former. The config.toml file can also remain as it is, and we could define the file inside the Docker Compose file to have a single configuration location. This shouldn't affect the current development environment in any way.

vroonhof:
OK

   - I don't intend to grow to into a fully supported end user business
   - Given the need to manually set up the flex export, run it and audit
   the output of the tax statement I doubt it will ever be easy
   - It is great to have docker config for people who like installing that
   way
   - it is not good to have a parallel command structure so if we are going
   to use docker, use ENTRYPOINT to just pass the normal interface along. I
   don't buy that environment variables are any easier.
   - The documentation has gotten confusing and the some of the
   feature example have gotten in the way (and are missing key arguments) of
   the main flow. Let's fix that separately
   - I have been weary of making the kursliste download implicit because it
   based on scraping their website, so it might break and it is impolite to do
   it too often.

I will leave the PR open for now for a docker script along the above lines.

anagno:
I also like the devcontainer approach and I think there might be 2 more alternatives:

* Just have a container where the dependencies are setup and a user can exec into it and be able to execute the commands as normal. E.g.:

```
docker pull image
docker run -it -v ${PWD}/necessary_files_in_host:/files image bash
opensteuerauszug process --importer ibkr <flex query xml file> ...
```

So no docker-compose etc. Just an alternative, instead of having to install manually python packages. 

* or just creating a devcontainer to simplify again the building procedure and running of the software

WDYT?

PR #330, opened 2026-03-30
chore: Migrate `UNINITIALIZED_QUANTITY` to optional

### LINKED ISSUE #314: constants.py: `UNINITIALIZED_QUANTITY` probably unsafe since we are now supporting short positions?
`UNINITIALIZED_QUANTITY = Decimal('-1')`

Is probably not that safe anymore - but  I have also not found anything broken because of this. We either should make it optional, or probably already 0 would be better?

pet-zh:
Resolves #314


vroonhof:
Still thinking abiut this one. This weakes the validation in the end. Ideaally pydantic world have multiple levels

pet-zh:
I am also not too sure this is the perfect solution. Another approach could be just using a _very_ high number here. For -1 chances are quite high it gets matched with a small short position.

vroonhof:
obseleted by PR378 which also adds a validation check. thanks for kicking this off.

PR #337, opened 2026-04-10
chore: Optimize imports

pet-zh:


src/opensteuerauszug/calculate/cleanup.py
```diff
@@ -3,17 +3,17 @@
 from typing import List, Optional, Dict, Any, cast, get_args
 from decimal import Decimal
 import logging
-from opensteuerauszug.model.ech0196 import (
-    SecurityTaxValue, TaxStatement, SecurityStock, BankAccountPayment, SecurityPayment,
-    Client, ClientNumber, CantonAbbreviation, LiabilityAccount, LiabilityAccountTaxValue,
+from ..model.ech0196 import (
```
vroonhof: Is it possible to not have the changes that add relative import paths (in particular the ones going to the product root still? I don't find them now read le

pet-zh: Would you prefer to have all imports chaged to absolute ones? We already have some relative ones in current main branch.

pet-zh: I have now replaced all `..` and `...` imports with absolute ones.

vroonhof:
Amazing how much junk got in because we don't enforce this with a linter..

PR #338, opened 2026-04-11
Add simple Flask based web app

pet-zh:


pet-zh:
This branch is currently deployed: https://opensteuerauszug.onrender.com/

src/opensteuerauszug/webapp/template.html
```diff
@@ -0,0 +1,272 @@
...
+          <p>This project is fully open source. Reports are very welcome to improve the software. Please report issues at <a href="https://github.com/vroonhof/opensteuerauszug/issues" target="_blank" rel="noopener noreferrer">GitHub Issues</a>.</p>
+        </div>
+      </details>
+      <details class="faq-item">
+        <summary>Are there alternatives?</summary>
+        <div class="faq-answer">
+          <p>Yes, there are paid alternatives available:</p>
+          <ul>
+            <li><a href="https://ibkr.datalevel.ch/" target="_blank" rel="noopener noreferrer">ibkr.datalevel.ch</a></li>
+            <li><a href="https://etaxify.ch/" target="_blank" rel="noopener noreferrer">etaxify.ch</a></li>
```
vroonhof, line 162: This almost the most interesting part of this change.  I didn't know about them, it is seems it is open season for estatement-as-a-service..

docs/user_guide.md
```diff
@@ -311,12 +311,62 @@ opensteuerauszug process input.xml --importer ibkr --tax-year 2025 \
...
+   ```
+
+2. Start:
+   ```bash
+   python src/opensteuerauszug/webapp/wsgi.py
+   ```
+   
+3. Access via your browser at http://localhost:10000
+
+### How to run the web app on Render.com (Free ("Hobby") Instance)
```
vroonhof: can we leave this part out (perhaps point it as an example). I think we should let the user do the due dilligenece of trusting the website hoster etc themselves.  I think running it like this completely changes the privacy and security posititioning of the project.

docs/user_guide.md
```diff
@@ -311,12 +311,62 @@ opensteuerauszug process input.xml --importer ibkr --tax-year 2025 \
 
 ---
 
+## Simple Web App
+
+This repository ships a minimal Flask web app. It supports IBKR XML uploads.
```
vroonhof, line 316: OOC, what is your main motivator for this. do you feel this provides a more friendly UI?

I am not 100% sure this is helpful if the user strays off the happypath in its current form.

docs/user_guide.md
```diff
@@ -311,12 +311,62 @@ opensteuerauszug process input.xml --importer ibkr --tax-year 2025 \
...
+
+### Setup
+
+1. Install dependencies:
+   ```bash
+   pip install ".[web]"
```
vroonhof, line 322: if we are targeting "simple users" perhaps an uv oneliner would be cool.

docs/user_guide.md
vroonhof: for instance I still think it not at all necessary for any of the tax data to ever leave the users browser.

vroonhof (commented):
Thank you for doing all this work.. However, I think any step in the direction of a service needs to be taken very very carefully.

I think having a web app implementaiton that is also labled very very clearly as an example is OK (perhaps it can be combined with the dockerization).

I didn't review the actual HTML in any depth,

docs/user_guide.md
pet-zh: The main motivation for adding this is not local usage. People able to run this via python / uv will also be able to use CLI.
Some (non tech) people I have suggested OSA told me that the do not feel comfortable installing python and asked if there are alternatives. IMO in the current statusy OSA is now quite ready for the broader audience. That also would help us to get even more testing and feedback.
If you do not want to merge this into main yet, it is also not an issue for me at all. Keep in mind that I also use this project as a smaller playground for trying out and learning new things while still working on something useful.

src/opensteuerauszug/webapp/template.html
pet-zh: Yes, just seen ads for this a few days ago at reddit. Looks like they are using the official PDF renderer. Would be interesting to know how the generate XML, or if the are even using existing tools around ;)

docs/user_guide.md
pet-zh: I tried to highlight to privacy topic. In addition, by default, client names and other optional attributes are not even leaving the browser.
I will try to address this even harder when I continue this PR.

vroonhof:
This is now obseleted because the AI agents can now almost one shot a pure standalone page: https://vroonhof.github.io/opensteuerauszug/

pet-zh:
@vroonhof Awesome! Your appoach is by far better. I did some basic tests and it seems to work. Let me fill some improvement ideas.

PR #343, opened 2026-04-12
Add option to use pixi instead of venv for dev work

hali8:
This adds an option which shouldn't cause any problems. (using pixi instead of venv)
My commits are a bit messy, sorry about that.
This is just to make it easier for me to add another broker (Fidelity Investments) and maybe for others to contribute

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ uv run --with git+https://github.com/vroonhof/opensteuerauszug.git opensteueraus
 
 Then continue with the [User Guide](docs/user_guide.md).
 
-### Development install [option 1] (contributors)
+### Development install [option 1] (contributors: using python venv)
 
 For development and tests, install from source:
 
@@ -134,7 +134,7 @@ For detailed documentation on available scripts, including the Kursliste filteri
 
 ### Setup
 
-To set up the development environment:
+To set up the development environment (venv):
 
 ```bash
 # Create and activate virtual environment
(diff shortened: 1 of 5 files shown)


hali8:
This can also just stay in my fork, of course!

scripts/setup_pixi.sh
```diff
@@ -0,0 +1,61 @@
...
+  PIXI_INSTALLED=true
+fi
+
+# check if pixi is already installed
+if [[ "$PIXI_INSTALLED" = false ]]; then
+  printf "No installed pixi found!\n"
+  printf -- "--> Installing pixi ..\n\n"
+  # installation for macOS or linux 
+  if [[ $OSTYPE == 'darwin'* || $OSTYPE == 'linux'* ]]; then
+    curl -fsSL https://pixi.sh/install.sh | sh
```
vroonhof: I don't think our scripts should do this behind the users back. Are you very likely to be a pixi user but not have it installed?

README.md
```diff
@@ -139,6 +146,16 @@ pip install -e ".[dev]"
 pip install git+https://github.com/vroonhof/pdf417decoder.git#subdirectory=python
 ```
 
+To set up the development environment (pixi):
```
vroonhof, line 148: Can we shorten this and just point pixi users to the e scripts README with one sentence?

vroonhof (changes requested):

README.md
hali8: attempted...

hali8:
Okay, I think I'm I'm done with this if you're happy with it

hali8:
@vroonhof I'm coming over to github from gitlab,  the interface shows me that you have requested  a change, but I cannot see what that change is.

pyproject.toml
```diff
@@ -49,8 +49,7 @@ dev = [
     "flake8>=6.0.0",
     "Flake8-pyproject",
     "freezegun>=1.5.0",
-    # TODO make these formal dependencies
```
vroonhof, line 52: I don't actually want to do this (for $REASONS )

Issue #342, opened 2026-04-12
Broker WH higher KL


hali8:
I'm trying to add a new (US based) broker. 
the broker withholding is much higher than the KurslisteTax withholding to Clarify:
[div_ok](https://github.com/vroonhof/opensteuerauszug/blob/6fc1cdab979ad3c7575caeeb01f620d9c97d08f1/src/opensteuerauszug/calculate/payment_reconciliation_calculator.py#L212) is fine but,
[w_ok](https://github.com/vroonhof/opensteuerauszug/blob/6fc1cdab979ad3c7575caeeb01f620d9c97d08f1/src/opensteuerauszug/calculate/payment_reconciliation_calculator.py#L218) fails with the broker withholding being higher than the KL withholding.
in order to attempt to properly handle this do you have a description of what [these](https://github.com/vroonhof/opensteuerauszug/blob/6fc1cdab979ad3c7575caeeb01f620d9c97d08f1/src/opensteuerauszug/calculate/payment_reconciliation_calculator.py#L61) signs are from ?

or maybe I should be concerned more with the _KurslisteAgg than the _BrokerAgg?

also the message [here](https://github.com/vroonhof/opensteuerauszug/blob/6fc1cdab979ad3c7575caeeb01f620d9c97d08f1/src/opensteuerauszug/calculate/payment_reconciliation_calculator.py#L229) is not correct:

In my case the withholding was above the kursliste value

vroonhof:
Hi @hali8. Thanks for contributing.

The signs in _BROKER_ABOVE_KURSLISTE_ALLOWLIST_SIGNS are for the case when
the dividend itself from the broker is higher. This occurs when the
dividend is tax-free from a Swiss perspective (e.g. return of capital or
income already taxed at the fund level, such as for direct real estate).

if w_ok fails due to high withholding tax from the broker I think there are
three likely cases

1. Kursliste withholding is zero because the kursliste is configured not
have claimable withholding. (this happens with funds that are missing the Q
flag) --> fix is to contact IC tax and ask them to add the flag.
2. Broker withholding for US security is 30% and kurliste tax is 15%. This
occurs because the broker is not applying W8-BEN, meaning you have to claim
the tax back from the IRS. This is not a bug per se but "unexpected" . We
may need a config setting "above-treaty-withholding" to suppress the
warning or add an explanation.
3. The security is located outside the US in a country that does not allow
W8-BEN style withholding reduction. e.g. The country is outside
[...]

PR #344, opened 2026-04-12
Above treaty withholding

hali8:
to close:
https://github.com/vroonhof/opensteuerauszug/issues/342

### DIFF
--- a/src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
+++ b/src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
@@ -225,12 +225,12 @@ class PaymentReconciliationCalculator:
                         allow_broker_above_kursliste or not security_has_sensitive_overwithholding
                     ),
                 )
-                if (broker_div_chf < kurs.dividend_chf  or broker_with_chf < kurs.withholding_chf):
+                if (broker_div_chf < kurs.dividend_chf  or (broker_with_chf is not None and broker_with_chf < kurs.withholding_chf)):
                     if not div_ok:
                         note = "Broker dividend is below Kursliste value beyond tolerance."
                     elif not w_ok:
                         note = "Broker withholding is below Kursliste value beyond tolerance."
-                elif (broker_div_chf > kurs.dividend_chf  or broker_with_chf > kurs.withholding_chf):
+                elif (broker_div_chf > kurs.dividend_chf  or (broker_with_chf is not None and broker_with_chf > kurs.withholding_chf)):
                     if not div_ok:
                         note = "Broker dividend is above Kursliste value beyond tolerance."
                     elif not w_ok and div_ok:
--- a/src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
+++ b/src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
@@ -73,9 +73,10 @@ class PaymentReconciliationCalculator:
         "kapitalgewinn",
     )
 
-    def __init__(self, tolerance_chf: Decimal = Decimal("0.05"), tolerance_frac: Decimal = Decimal("0.001")):
+    def __init__(self, tolerance_chf: Decimal = Decimal("0.05"), tolerance_frac: Decimal = Decimal("0.001"), allow_above_treaty_withholding: bool = False):
         self.tolerance_chf = tolerance_chf
         self.tolerance_frac = tolerance_frac
+        self.allow_above_treaty_withholding=allow_above_treaty_withholding
 
     def calculate(self, tax_statement: TaxStatement) -> TaxStatement:
         report = PaymentReconciliationReport()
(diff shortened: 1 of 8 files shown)


src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
```diff
@@ -223,10 +225,20 @@ def _reconcile_security(self, security: Security) -> List[PaymentReconciliationR
                         allow_broker_above_kursliste or not security_has_sensitive_overwithholding
                     ),
                 )
-                if not div_ok:
-                    note = "Broker dividend is below Kursliste value beyond tolerance."
-                elif not w_ok:
-                    note = "Broker withholding is below Kursliste value beyond tolerance."
+                if (broker_div_chf < kurs.dividend_chf  or (broker_with_chf is not None and broker_with_chf < kurs.withholding_chf)):
```
vroonhof: Rather than recomputing i think it would be cleaner to turn component_matches into a comparator (return low, equal, high)

src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
```diff
@@ -223,10 +225,20 @@ def _reconcile_security(self, security: Security) -> List[PaymentReconciliationR
...
+                    if not div_ok:
+                        note = "Broker dividend is below Kursliste value beyond tolerance."
+                    elif not w_ok:
+                        note = "Broker withholding is below Kursliste value beyond tolerance."
+                elif (broker_div_chf > kurs.dividend_chf  or (broker_with_chf is not None and broker_with_chf > kurs.withholding_chf)):
+                    if not div_ok:
+                        note = "Broker dividend is above Kursliste value beyond tolerance."
+                    elif not w_ok and div_ok:
+                        if (abs((broker_with_chf/kurs.withholding_chf)-2) < 0.01) and country == "US":
+                            note = ("Broker withholding is twice Kursliste value, check that your broker has a valid "
```
vroonhof: I think this ugly us specific logic is worth keeping.

src/opensteuerauszug/steuerauszug.py
```diff
@@ -113,6 +113,7 @@ def process(
     use_broker_withholding: UseBrokerWithholding = typer.Option(UseBrokerWithholding.CAP, "--use-broker-withholding", help="Control how broker withholding evidence is used: OFF disables adjustments, CAP (default) caps Kursliste (Q) withholding at the broker's effective level."),
     pre_amble: Optional[List[Path]] = typer.Option(None, "--pre-amble", help="List of PDF documents to add before the main steuerauszug."),
     post_amble: Optional[List[Path]] = typer.Option(None, "--post-amble", help="List of PDF documents to add after the main steuerauszug."),
+    allow_above_treaty_withholding: bool = typer.Option(False, "--above-treaty-withholding", help="Allow broker withholding > KL calculation"),
```
vroonhof: Would you be against making it a config option? These can also be at from the command line with --set and we can even lookup a broker specific value

vroonhof (changes requested):

src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
hali8: Is there a negation missing from your sentence ?

src/opensteuerauszug/steuerauszug.py
hali8: I'll try.

src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
vroonhof: sorry for the terse answer. What I meant was that special case logic like this in this place looks really unclean. But, given the high value of the "w8-ben not working" I am OK keeping it. assuming we do the comparator cleanup?

vroonhof:
I asked AI for help with my comments if you want to merge it in https://github.com/vroonhof/opensteuerauszug/pull/347 (a pure organic solution is fine too ;)). [it should have filed a PR against your branch but it made it against main[

hali8:
I didn't look at the AI generated one since it would probably be bd for my ego!
In implementing your comments I found that calculate_settings in steuerauszug.py couldn't be set from the command line, so I ended up making a few other changes. 
I think the new variable belongs in the calculate_settings even if that means it won't be a broker setting.
And my changes in test/{whatever} are prettyt hacky since I've never used unittest before

src/opensteuerauszug/calculate/payment_reconciliation_calculator.py
```diff
@@ -206,28 +207,36 @@ def _reconcile_security(self, security: Security) -> List[PaymentReconciliationR
...
-                if not div_ok:
-                    note = "Broker dividend is below Kursliste value beyond tolerance."
-                elif not w_ok:
-                    note = "Broker withholding is below Kursliste value beyond tolerance."
-                matched = div_ok and w_ok
+                if div_not_ok:
+                    div_note = f'Broker dividend differs from Kursliste value beyond tolerance. delta={div_not_ok} CHF.'
+                if w_not_ok:
+                    w_note = f'Broker withholding differs from Kursliste value beyond tolerance. delta={w_not_ok} CHF.'
+                note =' '.join([div_note, w_note])
```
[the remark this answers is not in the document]
vroonhof: fixed before merge

vroonhof:
made some fixes to the comments and type inconsistencies in the comparator.

I am having an attempt at rewriting the config resolve mechanism as you copying the pre-existing bad patterns makes it workse.

PR #352, opened 2026-04-15
Add DA1 country totals

### LINKED ISSUE #156: Consider rendering DA-1 subtotals by country

Oppizz:
Add total per country to DA1 section in PDF
Resolves #156


vroonhof:
Thank you very much!

PR #357, opened 2026-04-19
Basic support for stock/future options

Oppizz:
Basic support to import statements with options
- prevent error with zero price (e.g. expirations)
- import opening balance even if negative (import of generated statement works in eFisc TG)
- includes real live test data options_future_2025.xml


vroonhof:
> * import opening balance even if negative (import of generated statement works in eFisc TG)

What does eFisc actually do with negative end of year balances for options?

vroonhof:
Merging this in, thanks for contributing. I am still unclear on what the official way to handle these negative options is, but if we are going to enforce some policy or warning this probably better happens in a later state. Given  we allow negative end balances we should allow negative start balances.

Oppizz:
Import works fine and it shows as negative end of year balance, but it wouldn't let me enter the same value manually.
However, negative opening balance can be entered manually... 
image

Oppizz:
Another interesting fact: schema eCH-0196 allows negative values, not sure if that's intentionally or accidentally...

PR #361, opened 2026-04-19
Apply option multiplier to OpenPosition

Oppizz:
Mark price should also reflect the multiplier


vroonhof (approved):
it is interesting that the the standard does not have a more meaningful way to support this, as you would want to make it easy to compare the options prices with the market convention. (

PR #365, opened 2026-04-21
Set dividend 0 for short stock

Oppizz:
Negative dividends (payment in lieu) due to short stock on record date are not tax deductible and cannot be treated as fees or interest payment (according to various court rulings). 
Set dividend to 0 but leave quantity and amount per unit as is (tax software efisc TG does the same when manually entering negative qty)


vroonhof (approved):
Are we actually allowing short stock at this point or is there more logic coming? 

(if so would be good to have this be a config flag because negative balances are possibly also an indication of an importer gone wrong.

Oppizz:
This only considers quantity on record date for the payment. 
Short stock at start or end of year is not part of this.

PR #370, opened 2026-04-23
Shared importers/common/postprocess stage + port IBKR and Schwab onto it

vroonhof:
## Summary

Introduces the shared post-processing stage in `importers/common/postprocess.py` and ports all three importers (IBKR, Fidelity-ready in a follow-up, Schwab) to it. Previously stacked on top of the postprocess-stage branch; now rebased directly on top of `main` (which has the Fidelity importer but not yet the shared postprocess stage — the Fidelity migration follows in a stacked PR).

Commit breakdown:

1. **`Add shared post-processing stage in importers/common`** — new `importers/common/postprocess.py` with `augment_list_of_securities(statement, positions, ...)` and `augment_list_of_bank_accounts(statement, entries)`, a `PositionHints` value object for per-position category / country / short-position / skip flags, a `CashAccountEntry` record for the bank-accounts side, and a `fold_cash_payments(seeds, processed_cash_positions)` helper. API takes a partially-filled `TaxStatement` so importer-specific scalars (institution, client, canton, period dates) live on the statement itself. 8 new unit tests.
2. **`Migrate IBKR importer onto the shared post-processing stage`** — IBKR `−276/+86` lines. Translates `(asset_cat, sub_category)` → `PositionHints` via a local closure. `CashAccountEntry` seeds still built from IBKR's `CashReport` + `AccountInformation`.
3. **`Let CashAccountEntry.number stay None instead of auto-defaulting`** — required for Schwab awards (synthetic `f"{id}-{curr}"` would exceed the 32-char eCH limit). IBKR seeds now set `number=…` explicitly.
4. **`Migrate Schwab importer onto the shared post-processing stage`** — Schwab `+239/−195`. `position_map` aggregator is split into the shared TypedDict shapes; Schwab-specific display names (`AWARDS`, `...<id>` fallback, `Equity Awards <cash_id>`, `(Unsettled)` suffix) live in two pure helpers (`_resolve_security_depot_display_name`, `_resolve_cash_account_identity`). T+1 settlement splitting (`split_unsettled_cash`) stays in Schwab. Dead helpers removed.
5. **`Drop the three legacy Schwab converter functions`** — `convert_security_positions_to_list_of_securities`, `convert_cash_positions_to_list_of_bank_accounts` and `create_tax_statement_from_positions` were only reachable from their own unit tests after the migration. Extracted `_pick_primary_client_number` and `_schwab_security_display_name` so the rules they embodied remain testable as small pure helpers; redirected all relevant tests at those helpers. `tests/importers/schwab/test_conversion_helpers.py` removed. Net `−333` lines.
6. **`Add assume_zero_if_no_balances + aggregate_same_day_mutations knobs`** — addresses both PR review comments:
   - `assume_zero_if_no_balances: bool = False` on `augment_list_of_securities` so mutation-only histories reconcile to the correct net closing balance. Schwab passes `True` (matches legacy `_reconcile_and_ensure_boundary_stocks_for_position`). Codex P1 fix.
   - `aggregate_same_day_mutations: bool = True` so callers can opt out of merging same-day rows. Schwab passes `False` — awards accounts carry per-vesting text on the `name` field that the default aggregator would silently drop. @vroonhof fix.

## Test plan

- [x] `python -m pytest` — 753 passed, 6 skipped (vs 750 before the rebase; the `+3` is from main's Fidelity tests)
- [x] `python -m pytest tests/importers/schwab/` — 127 passed, 1 skipped
- [x] `python -m pytest tests/importers/common/test_postprocess.py` — 12 passed
- [x] `python -m pytest tests/importers/ibkr/` — full IBKR suite including integration tests passes against the existing goldens
- [x] Schwab `test_mutation_only_security_history_does_not_fail_initial_reconciliation` keeps passing (confirms the new `assume_zero_if_no_balances=True` path gives the correct net closing balance)
- [ ] Manual check on a real Schwab awards statement to confirm per-vesting descriptions are preserved on the rendered PDF.

## Stacked PR

The Fidelity unification — migrating main's Fidelity importer onto the same shared postprocess — follows in a separate PR stacked on top of this one.

https://claude.ai/code/session_011ASnWNx2WUUYzmJu3ErHM1


src/opensteuerauszug/importers/schwab/schwab_importer.py
```diff
@@ -173,163 +241,6 @@ def __init__(self,
...
-            cash_account_id=pos_obj.cash_account_id,
-            is_unsettled_balance=True,
-        )
-        logger.info(
-            f"[{unsettled_pos.get_processing_identifier()}] Created unsettled cash account "
-            f"with {len(unsettled_stocks)} trade(s) totalling {total_unsettled} {currency}."
-        )
-        return unsettled_pos, all_stocks, []
-
     def import_files(self, filenames: List[str]) -> TaxStatement:
```
vroonhof, line 280: after this change we now have the behavior that the shared postprocessor defaults to combine different trades on the same day into one. For schwab (and in particular awards) account, this is the wrong behavior as the description field contains key vesting information (and just re-using the description from the first one is actively misleading), we should add the ability to not do this recombination,

src/opensteuerauszug/importers/schwab/schwab_importer.py
```diff
@@ -524,56 +435,189 @@ def import_files(self, filenames: List[str]) -> TaxStatement:
...
+                tax_statement.client = [
+                    Client(clientNumber=ClientNumber(setting.account_number))
+                ]
+                break
+
+        augment_list_of_securities(
+            tax_statement,
+            processed_security_positions,
+            name_registry=name_registry,
+            hints_for=lambda _sp: PositionHints(
```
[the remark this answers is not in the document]
vroonhof: Good catch — you're right that without `assume_zero_if_no_balances=True`, a mutation-only symbol with a non-zero net (e.g. a standalone buy of +10 that never appears on a position snapshot) would have its closing balance collapsed to zero.

Fixed in aca0be0: added an `assume_zero_if_no_balances` parameter on `augment_list_of_securities` (default `False` to preserve IBKR / Fidelity behaviour, since those brokers always emit OpenPositions). Schwab passes `True`, which routes both `synthesize_position_at_date` calls through the same path the legacy `_reconcile_and_ensure_boundary_stocks_for_position` used.

New pin at `tests/importers/common/test_postprocess.py::test_augment_securities_assume_zero_walks_mutations_without_balances` locks in the mutation-only-buy → closing=10 case.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/opensteuerauszug/importers/schwab/schwab_importer.py
vroonhof: Agreed — "first description wins" is actively misleading for awards because each vesting has its own meaningful text.

Fixed in aca0be0: added `aggregate_same_day_mutations: bool = True` to `augment_list_of_securities`. Schwab passes `False`, so every extractor row becomes its own `SecurityStock` entry, including cases where two grants vest on the same day. IBKR / Fidelity use the default `True` to preserve the broker-lot fold they rely on.

Pinned by two new postprocess tests:
- `test_augment_securities_preserves_same_day_mutations_when_aggregation_off` — two same-day same-sign vesting rows keep their individual `name` text.
- `test_augment_securities_aggregates_same_day_mutations_by_default` — same-day same-sign rows still fold when the flag is left on (regression guard for IBKR / Fidelity).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #372, opened 2026-04-24
Add other importers to init

hali8:
In case this needs to be symmetric / consistent for all importers ...


vroonhof:
I'm still not sure iif these add vakue but consistency is definably good

PR #374, opened 2026-04-24
fix: allow negative balances to pass through importer for all asset types

GitHubEtienne:
## Summary

Fixes the crash on short stock positions (STK) such as CVNA or TSLA held short at year-end.

The importer previously clamped negative computed balances to zero and raised `ValueError` for any asset category other than OPT/FOP. This prevented processing of accounts with short stock positions.

Per the preferred architecture (enforcement at the calculation stage, not the importer), the importer now passes negative balances through for all asset types with a debug log. The calculation stage already handles them correctly: `price * quantity` with a negative quantity produces a negative CHF value in the Wertschriften section.

## Changes

- `src/opensteuerauszug/importers/ibkr/ibkr_importer.py`: remove clamp-to-zero and `ValueError` for negative balances; replace with a `logger.debug` that notes the negative and defers to the calculation stage

## Real-world validation

- Short positions: CVNA (−1), TSLA (−5) at 31.12.2025
- Generated XML imported into eTax.Zug (Kanton Zug) without error
- Negative Wertschriften values accepted by cantonal software
- Net wealth matched IBKR NAV within CHF 150 (interest accruals + FX rounding)

Relates to: #309

### DIFF
--- a/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
+++ b/src/opensteuerauszug/importers/ibkr/ibkr_importer.py
@@ -1048,19 +1048,14 @@ class IbkrImporter:
                 opening_balance = start_pos.quantity
             else:
                 tentative_opening = closing_balance - trades_quantity_total
-                opening_balance = tentative_opening if tentative_opening >= 0 or asset_cat in ["OPT", "FOP"] else Decimal("0")
+                opening_balance = tentative_opening
 
             if opening_balance < 0 or closing_balance < 0:
-                if (asset_cat == "OPT" or asset_cat == "FOP") and (sub_category == "C" or sub_category == "P"):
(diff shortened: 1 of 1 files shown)


vroonhof:
Obsoleted by PR#379 we can then add a flag in one place to allow short positions.

GitHubEtienne:
Thanks — happy to close this in favour of #379, which handles it more cleanly at the right layer. Also noted your feedback on #375 and #376 — makes sense to let those land through the fill-in mode path rather than as separate patches.

vroonhof:
Note that 379 moves the logic but does not actually allow negative values.
So it is cleaner but not a full solution for you. I also asked the Zürich
tax office but have yet to get a response

Issue #261, opened 2026-03-04
IBKR: Short Positions


thbch:
I stumbled over this project last weekend and it would be really great to have solution like this! So thank you for your efforts! 

I tried to generate two different years and stumbled over different problems - so this one here seems related to short positions (yes I have them, short options). However based on the message I assume there is something going wrong where it shouldnt.

*********

WARNING:opensteuerauszug.importers.ibkr.ibkr_importer:Negative balance computed for security 746386368 with OPT/C. In case you expect short positions, this is fine. Otherwise, please report this to the developers for further investigation. (start 0, end -2)
Error during phase import: Negative balance computed for security 666235415. In case you expect short positions, please report this to the developers for further investigation. (start 0, end -1)
Stack trace:
Traceback (most recent call last):
  File "D:\git\opensteuerauszug\.venv\Lib\site-packages\opensteuerauszug\steuerauszug.py", line 351, in main
    statement = ibkr_importer.import_files([str(input_file)])
  File "D:\git\opensteuerauszug\.venv\Lib\site-packages\opensteuerauszug\importers\ibkr\ibkr_importer.py", line 895, in import_files
    raise ValueError(
    ...<2 lines>...
    )
ValueError: Negative balance computed for security 666235415. In case you expect short positions, please report this to the developers for further investigation. (start 0, end -1)

*********

What kind of information do you need to investigate on this one?

vroonhof:
What is most useful is having a small anomized bit of flex xml that shows the problem. What type of security is 666235415?

One issues is that it not entirely ckear how short positions should be handled in the output. The standard is vague. How would you have entered them manually? We should also check we are not booking negatieve income etc

thbch:
I extracted both positions (746386368, 666235415) from the export. This is all related to options. 

For 666235415, the option was opened on 8.11.2024 and then closed on 27.2.2025 (so only one transaction in 2025).

For 746386368 (and 746386727) , these positions were opened on 30.12.2024 and closed on 22.1.2025.

So it seems the problems arise due to positions held over year end. 

[option-execution.xml](https://github.com/user-attachments/files/25828247/option-execution.xml)

pet-zh:
@thbch Thanks for sharing details.

I have not tested your samples yet, but I assume the ones with `subCategory="C"` should work, but the one with `subCategory="P"` not. Adding sub category `P` would be an easy change to not make this import fail anymore.

1. Do you have an idea what those sub categories mean?
[...]

PR #375, opened 2026-04-24
fix: fall back to broker balance for open options not in Kursliste

GitHubEtienne:
## Summary

US equity options have no ISIN and are not listed in the ESTV Kursliste. Previously, any open option position at year-end was marked `undefined=1` and valued at CHF 0, silently understating taxable wealth.

This adds a fallback path: when an option has a non-zero closing balance and is not found in the Kursliste, the broker's reported `positionValue` is converted to CHF using the SNB year-end exchange rate and used as the `SecurityTaxValue`. If `positionValue` is absent from the flex query, an explicit `logger.warning` is emitted and `undefined=1` is set — the user is notified rather than silently getting CHF 0.

**Requirement:** The IBKR flex query must include the `positionValue` field.

## Changes

- `src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py`:
  - Add `_current_security_is_missing_open_option` state flag (init + reset in `_handle_Security`)
  - In `_handle_Security`: add `elif is_option and closing_balance != 0` branch to set the flag
  - Also clean up the existing `if closing_balance == 0` redundant guard (always true in that branch)
  - In `_handle_SecurityTaxValue`: add fallback handler that converts broker balance to CHF and populates `unitPrice`, `value`, `balance`, `exchangeRate`, `balanceCurrency`

## Real-world validation

- 17 short option positions open at 31.12.2025 (various US equity underlyings)
- Previously valued at CHF 0 total (undefined=1); with this fix correctly valued at ~CHF 26,000
- Generated XML imported into eTax.Zug (Kanton Zug) without error

Relates to: #261

### DIFF
--- a/src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
+++ b/src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
@@ -105,6 +105,7 @@ class KurslisteTaxValueCalculator(MinimalTaxValueCalculator):
         self.render_language = render_language
         self._current_kursliste_security = None
         self._current_security_is_zero_balance_option = False
+        self._current_security_is_missing_open_option = False
         self._missing_kursliste_entries = []
         self._stock_split_warnings: List[dict] = []
         self._previous_year_exdate_warnings = []
@@ -164,6 +165,7 @@ class KurslisteTaxValueCalculator(MinimalTaxValueCalculator):
     def _handle_Security(self, security: Security, path_prefix: str) -> None:
         self._current_kursliste_security = None
         self._current_security_is_zero_balance_option = False
+        self._current_security_is_missing_open_option = False
 
         if not self.kursliste_manager:
             super()._handle_Security(security, path_prefix)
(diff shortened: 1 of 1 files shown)


vroonhof:
I think there are multiple issues here. So i don't think this is the right
fix.

A. It should never have failed silently. (Perhaps this is related to no
isin). A warning should appear in the output and more importantly in the
pdf.

B. You are not the first to suggest this. However the behaviour you want is
tax-calculation-mode=fill-in

(There is probably room for an intermediate setting that only fills in safe
eoy values for stuff that will never be on the kurskiste)

B2. When we do this we should adapt the cover information that says only
kursliste values are used

C. The computation from the broker value should already have happened
before in cleanup (and then not cleared) instead of Computing again later

I am at my phone right now and cavity y look at the actual code

src/opensteuerauszug/calculate/kursliste_tax_value_calculator.py
```diff
@@ -273,6 +283,29 @@ def _handle_SecurityTaxValue(self, sec_tax_value: SecurityTaxValue, path_prefix:
                 self._set_field_value(sec_tax_value, "value", Decimal("0"), path_prefix)
                 self._set_field_value(sec_tax_value, "exchangeRate", rate, path_prefix)
             return
+        elif self._current_security_is_missing_open_option:
+            # Open option not in Kursliste: use broker's reported position value converted to CHF.
+            # Requires positionValue to be included in the IBKR flex query.
+            if sec_tax_value.balance is not None and sec_tax_value.balanceCurrency and sec_tax_value.referenceDate:
```
vroonhof, line 289: this computation already is in MinimalTaxValueCalculator so the right thing to is to fall through the base implementation.

vroonhof: Actually the fallback wasn't working correctly. I have now moved it to the Fill-in calculator which should now work. 

The TODO is now to make a new in betewen setting that does relatively safe fill ins like this one by default.

vroonhof:
I think this PR now mostly obselete. Lets track actually making options work by default as Issues 261

PR #420, opened 2026-07-20
fix: preserve all DEGIRO cash balances

### LINKED ISSUE #396: [DEGIRO] Keep all cash rows instead of overwriting DEGIRO cash balance**
**<sub><sub>P1 Badge</sub></sub>  Keep all cash rows instead of overwriting DEGIRO cash balance**

The importer stores cash in single scalars (`cash_balance`/`cash_currency`) and overwrites them for every cash row in `Portfolio.csv`, then emits only one bank account. If a DEGIRO portfolio contains multiple cash currencies (e.g., CHF and USD/EUR balances), all but the last row are dropped from the generated statement, which understates assets and produces incorrect tax data.

Useful? React with 👍 / 👎.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/vroonhof/opensteuerauszug/pull/395#discussion_r3182322530_

Sanjays2402:
Fixes #396

DEGIRO portfolio imports previously retained only the final cash row, dropping balances in earlier currencies. This collects every cash row and emits a separate bank account for each currency.

Tested with the full DEGIRO importer suite (97 tests), including a new CHF/USD regression.

### DIFF
--- a/src/opensteuerauszug/importers/degiro/degiro_importer.py
+++ b/src/opensteuerauszug/importers/degiro/degiro_importer.py
@@ -26,7 +26,7 @@ import re
 from collections import defaultdict
 from datetime import date, timedelta
 from decimal import Decimal
-from typing import Dict, List, Optional
+from typing import Dict, List
 
 from opensteuerauszug.config.models import DegiroAccountSettings
 from opensteuerauszug.importers.common import (
@@ -139,8 +139,7 @@ class DegiroImporter:
         )
 
         # Step 1 – Seed closing balances from Portfolio.csv
-        cash_balance: Optional[Decimal] = None
-        cash_currency: str = "CHF"
+        cash_balances: List[tuple[Decimal, str]] = []
 
         end_plus_one = self.period_to + timedelta(days=1)
 
@@ -149,8 +148,7 @@ class DegiroImporter:
         isin_to_entry: Dict[str, PortfolioEntry] = {}
         for entry in portfolio_entries:
             if entry.is_cash:
-                cash_balance = entry.local_amount
-                cash_currency = entry.local_currency or "CHF"
+                cash_balances.append((entry.local_amount, entry.local_currency or "CHF"))
                 continue
             if not _valid_isin(entry.isin):
                 logger.debug("Skipping portfolio entry with non-ISIN: %s", entry.isin)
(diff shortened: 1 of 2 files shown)


vroonhof:
Hi Sanjay. What evidence do have that degiro actually uses this format. I have not found any real world examples with multiple currencies.

In general what brings you to contribute here to swiss tax software and/or have a degiro account? Could it be you are my first full AI contributor, if so please disclose that?

Sanjays2402:
I cannot substantiate the multiple-currency assumption with a real DEGIRO export; the regression fixture in this PR is synthetic. Without evidence that DEGIRO actually emits multiple currency cash rows in this format, the premise is too weak to ask you to merge this. Please close this PR; if I can obtain a genuine anonymized export showing the case, I will reopen the investigation with that evidence.

zarubaf:
I encountered this issue with my actual DEGIRO Portfolio.csv export for year-end 2025. It contains separate CHF and USD cash rows:

```csv
Produkt,Symbol/ISIN,Anzahl,Schlußkurs,Wert,,Wert in CHF
CASH & CASH FUND & FTX CASH (CHF),,,,CHF,500.00,500.00
CASH & CASH FUND & FTX CASH (USD),,,,USD,200.00,158.00
```

The header and row structure are preserved from the real export; amounts have been replaced for privacy and securities omitted.

This supports the premise of this PR: retaining only the final cash row drops the earlier CHF balance. Could this evidence unblock the review?

Disclosure: I used Codex to investigate the issue and prepare this comment.

Sanjays2402:
@zarubaf I have fixed the issue, please check

vroonhof:
Reviewed this against @zarubaf's real export, and added the missing end-to-end coverage on branch [`claude/degiro-export-e2e-test-7raezn`](https://github.com/vroonhof/opensteuerauszug/tree/claude/degiro-export-e2e-test-7raezn) (both of @Sanjays2402's commits cherry-picked, authorship preserved).

**Evidence check.** I turned the reported rows into a committed sample (`tests/samples/import/degiro/de_multi_currency_cash/`, header and row structure as reported, anonymized amounts, hand-written `Account.csv` consistent with the snapshot). On `main` the import yields exactly the reported data loss — only `[('USD', 200.00)]`, the CHF 500.00 row silently gone. With the fix both survive, and the full CLI run renders a statement with `12345678-CHF` (500) and `12345678-USD` (200 → CHF 158.45 at the mini-Kursliste rate), still XSD-valid. That test (`tests/test_integration_degiro_2025.py`) fails without the fix, so the premise is now covered by a regression test rather than a synthetic fixture.

**On the diff itself** — the approach looks right to me:
* dropping the `Optional` import is safe (no other use in the file);
* with no cash rows at all, `augment_list_of_bank_accounts(statement, [])` leaves `listOfBankAccounts` as `None`, i.e. unchanged from the old `if cash_balance is not None` guard;
* account ordering follows Portfolio.csv order, which is deterministic.

Two small things I added on top: a comment recording that a listed-but-empty wallet deliberately still produces a zero-balance account (the `-200.00` parametrization documents this), and a warning instead of silently folding a cash row with no currency column into CHF.

**Separate defect found while writing the e2e test** (fixed in its own commit on that branch): `ConfigManager` had no `degiro` branch, so the `[brokers.degiro.accounts.*]` section documented in the DEGIRO guide was dropped with a warning. Every DEGIRO run therefore fell back to the `DEGIRO` depot id and the `NOIDENTI` placeholder client number — i.e. the bank accounts this PR emits were numbered `DEGIRO-CHF`/`DEGIRO-USD` rather than by the configured account number.

---
_Generated by [Claude Code](https://claude.ai/code)_

vroonhof:
Superseded by PR432.

PR #434, opened 2026-09-16
fix(degiro): recognize newer German export formats and fix dividend/withholding reconciliation

Carismarkus:
## Summary

  While processing a real 2026 DEGIRO German-language export, two independent bugs
  surfaced:

  1. **New German export row formats not recognized** (first commit): buy/sell rows
     switched from `Kauf 60 iShares@20.08 EUR (ISIN)` to `Kauf 3 zu je 64,91 EUR
     (ISIN)`, which affected *every* trade row in a current German export — all
     silently skipped, no trades imported at all. Three other row types
     (`Einrichtung von Handelsmodalitäten ...`, `Währungswechsel
     (Einbuchung/Ausbuchung)`, `Geldmarktfonds Kompensation`) were also
     unclassified and caused a hard `NotImplementedError` abort.

  2. **Dividend + withholding tax merged into one payment breaks reconciliation**
     (second commit): `_process_dividend()` merged the dividend amount and
     withholding tax onto a single `SecurityPayment`, but
     `PaymentReconciliationCalculator._accumulate_broker()` expects them as two
     separate payments (the convention the IBKR/Fidelity importers already
     follow) — causing every DEGIRO dividend with withholding tax to show a false
     "Broker div: None" mismatch in the reconciliation report.

  Both are fixed with minimal, targeted changes; see the individual commit
  messages for full detail. Each is covered by new/extended unit tests.

  ## Test plan

  - `pytest tests/importers/degiro/` — 109 passed
  - Full suite: `pytest tests/` — same result as `main` (2 pre-existing, unrelated
    Windows-only failures in fidelity/build_web_app tests)
  - Verified end-to-end against a real 2026 DEGIRO export: import now completes
    without warnings/aborts, and payment reconciliation correctly matches
    dividends that were previously flagged as false mismatches
  - `black` / `flake8` clean
  
  AI used: Claude Sonnet 5


vroonhof (approved):
Thank you. It is bit sad DEGIRO cannot produce something more robust.

